Issue #6, opened 2020-07-21
Add Array Destructuring to JS3-3 and Remove Classes


ChrisOwen101:
**Which module(s) and week(s) does this change effect?**
Module: JavaScript Core 3
Week 3

**What is the work that needs to be done?**

Removing Classes
1. Remove content about `Class`es from JS3-3
2. Remove content about `Class`es from JS3-3 homework

Adding Array Destructuring
1. Write content that properly explain the usages of array destructing and why it will be helpful.
2. Write several in-class exercises which teach the students how to use array destructuring (consider using [Faded Examples](https://www.researchgate.net/figure/Sample-faded-worked-example_fig2_282434634))
3. Add homework tasks that ask the student to write some array destructuring code
4. Add homework tasks that ask the student to convert existing code to using array destructuring

**Why is this work important to do?**

React has now moved from using classes to using `hooks` which means that JS Classes now have far less importance in our course. For this reason we should update our curriculum so that our students are better prepared for React.

**Additional context**

**Who might need to know about this change?**

@ChrisOwen101

40thieves:
This also has down-stream effects on React as we did have a (short) section when teaching `useState` to cover destructuring. See the bottom of [this section](https://syllabus.codeyourfuture.io/react/week-2/lesson#using-usestate).

It's on my list to remove this section (because we decided to move it earlier, as discussed here). I'll coordinate with @ChrisOwen101, so that we don't accidentally drop it for a class.

40thieves:
Update: I've left the link to the Wes Bos post about destructuring in my [React update PR](https://github.com/CodeYourFuture/syllabus/pull/36), so that Birmingham can include it there. The section is somewhat light on the concept - the idea was to rely on the post for the content, to try keep the focus on React.

The consensus from the London post-module retro was that it would be better to do more practice of the concept in isolation, before combining with `useState`.

Looking at it again now, I think I'd be in favour of keeping some recap of the concept in the React module. But I'll hold off on making that change until this one is decided upon.

nbogie:
@40thieves yeah it would be good to keep relevant recap material for JS pre-reqs linked in the relevant spot in react module, and to leave it to the discretion of the city teachers to gauge whether it is needed in their particular case.

PR #36, opened 2020-07-29
React hooks updates

alismith-dev:
# What does this change?

Module: React
Week(s): Weeks 1, 2 & 3

## Description

Updates the Hooks content based to make the changes suggested in the volunteers retro in London after teaching class 6.

[Rendered version - week 1](https://github.com/CodeYourFuture/syllabus/blob/react-hooks-updates/docs/react/week-1/lesson.md)
[Rendered version - week 2](https://github.com/CodeYourFuture/syllabus/blob/react-hooks-updates/docs/react/week-2/lesson.md)
[Rendered version - week 3](https://github.com/CodeYourFuture/syllabus/blob/react-hooks-updates/docs/react/week-3/lesson.md)

### Summary of changes

- Converted week 1 to use table formatting for exercises
- Added exercise time estimates
- Added more examples & exercise showing the `useState` naming convention
- Adjusted state exercises to make more sense (mostly unified into a single complete exercise)
- Added section on "complex" types in state (e.g. arrays & objects)
- Changed multiple state exercise to be more realistic
- Added props vs state exercise
- **Moved `useEffect` content to week 3**
- Added explanation of why `fetch` directly inline in a component won't work
- Added importing `useEffect` section
- Added tip to write `useEffect` structure before the callback (hopefully will avoid accidental rate limiting in the future)
- Fixed `useEffect` exercise wording
  - Mostly by adding a new component, instead of updating an existing component, which should also help reduce problems with students not completing previous exercises
- Fixed some confusing prop naming (the state and prop were named the same)
- Added single form element exercise
- Moved complex multiple form fields pattern to further reading (on a separate page)
- Moved container components section to further reading (on a separate page)

~Note that this PR contains some exercise changes, which will invalidate the completed exercises [here](https://github.com/CodeYourFuture/react-pokedex-exercises-solution). But I have made a note to update those.~ I have now prepared these changes, but I'll hold off pushing them until this is merged.

### Things NOT changed

Because Birmingham haven't been through the module yet, we decided to keep the array destructuring content within the React module. Once they have gone through the module, in can be removed as discussed in https://github.com/CodeYourFuture/syllabus/issues/6.

@ChrisOwen101 
@LucyMac 
@llh1 
@anthonytranDev


40thieves:
I'm a bit unsure how to scope this change <sup>[can you tell I'm doing more management? 😂 ]</sup>

I'd definitely like to do some further updates, but I'm starting to think that this is enough of a diff to be reviewable, and do some follow-ups later.

docs/react/week-3/lesson.md
```diff
@@ -222,7 +360,7 @@ Phew! That was a lot of work just to render an image! But we're not quite done y
         ...
         <button onClick={handle2020Click}>Fetch image for 2020</button>
         ...
-        {date ? <MartianImageFetcher={date} /> : null}
+        {date ? <MartianImageFetcher photoDate={date} /> : null}
```
illicitonion, line 372: I'm worried that, having not explained any of this (i.e. the dependencies array, or even what a dependency is), bullet point 18 is confusing, and may leave students feeling like this is more complicated than it is just because they're not familiar with the terminology - particularly as the phrasing of the bullet point suggests to me that the reader _should_ already know this, but we didn't explain any of these concepts to them yet.

Perhaps we could add something to the end of the previous "The timeline of this component" para, like "When the response is received, update the state and re-render so that we can show the data on-screen. The way we've called it, `useEffect` _only_ calls its callback function the first time a component renders - when we do the re-render to show the data on-screen, we will _not_ do another `fetch`."

Then we could rephrase this bullet point as:
This time, we do **not** queue an effect. React doesn't know that it should - remember, the point of `useEffect` is to not call the callback function every time the component renders, so we only queued the effect on the **first** render!

docs/react/week-3/lesson.md
```diff
@@ -263,90 +406,47 @@ function MartianImageFetcher(props) {
...
     ...
-  }, [props.date]);
+  }, [props.photoDate]);
 
   ...
 }
 ```
 
-Now when the `date` prop changes, React knows that the effect must be run again, this time with the 2020 data. Because of this behaviour, the second argument to `useEffect` is called the _dependencies argument_. We use it whenever we have something in our effect that _depends_ on a variable outside of the effect function.
+Now when the `photoDate` prop changes, React knows that the effect must be run again, this time with the 2020 data. Because of this behaviour, the second argument to `useEffect` is called the _dependencies argument_. We use it whenever we have something in our effect that _depends_ on a variable outside of the effect function.
```
illicitonion:
```suggestion
Now when the `photoDate` prop changes, React knows that the effect must be run again, this time with the 2020 date. Because of this behaviour, the second argument to `useEffect` is called the _dependencies argument_. We use it whenever we have something in our effect function that _depends_ on a variable outside of the effect function.
```

docs/react/week-3/lesson.md
```diff
@@ -379,32 +479,44 @@ If we didn't call `setReminder` in the `handleChange` function, then the input's
...
+| 6. Set the `value` attribute of the `<input>` to the `city` state.                                                                                                                                      |
+| 7. Create a function within the component called `updateCity`. Pass this function to the `onChange` event handler.                                                                                      |
+| 8. Change the `updateCity` component so that it has a parameter named `event`.                                                                                                                          |
+| 9. Add a `console.log` to inspect the value of `event.target.value`. What happens when you type in the input?                                                                                           |
+| 10. Using `setCity`, update the city state to what what is typed in the input box.                                                                                                                      |
```
illicitonion:
```suggestion
| 10. Using `setCity`, update the city state to what was typed in the input box.                                                                                                                      |
```

docs/react/week-3/lesson.md
```diff
@@ -444,81 +556,49 @@ function CreateAccountForm() {
 }
 ```
 
-We now have three different inputs named `username`, `email` and `password`, and we keep each entered value in state as a field in an object. The method `handleChange` is reused to keep track of changes for **all** the values. The trick here is to use the `name` of the `<input>` element to update the corresponding state.
+We now have three different inputs named `username`, `email` and `password`, and we keep each entered value in state as a field in an object.
```
illicitonion: I don't see an object? It looks like we're using three state hooks in a function? Maybe reword?

docs/react/week-3/lesson.md
```diff
@@ -444,81 +556,49 @@ function CreateAccountForm() {
...
 
-#### Updating an Object in State
+The `handleChange` function is reused to keep track of changes for **all** the values. To be able to tell which `<input>` is being updated, we check the `event.target.name` field. This corresponds to the `name` attribute on the `<input>` (e.g. `<input name="username">`).
```
illicitonion:
```suggestion
The `handleChange` function is reused to keep track of changes for **all** the form fields. To be able to tell which `<input>` is being updated, we check the `event.target.name` field. This corresponds to the `name` attribute on the `<input>` (e.g. `<input name="username">`).
```

docs/react/week-3/lesson.md
```diff
@@ -444,81 +556,49 @@ function CreateAccountForm() {
...
-};
-
-const sherlockAfterMoving = {
-  ...sherlock,
-  address: "221b Baker Street",
-};
-
-console.log(sherlock);
-console.log(sherlockAfterMoving);
+if (event.target.name === "username") {
```
illicitonion: Maybe this lesson is already long enough, but this could be a really nice opportunity to show a contrasting implementation where we use arrow functions as event handlers, and to remind the students once again about needing function references. Something like:

```jsx
<input onClick={event => setUsername(event.target.value)} />
<input onClick={event => setEmail(event.target.value)} />
<input onClick={event => setPassword(event.target.value)} />
```

This is something I've seen them try to do, and get stuck on a lot. Very frequently they try to write;
```jsx
<button onClick={incrementCounter()} />
```
and get confused when it doesn't work.

The fact that the single event handler function is a bit involved and fiddly, and these little arrow functions are concise, leads to a nice contrasting (and then we could maybe say something like "But if you're doing some validation or something, maybe the single function is a better approach because you can centralise logic"). Contrasting approaches, and thinking about why one is better than the other, isn't something we do much of, and would be great to do more of :)

docs/react/week-3/learning-objectives.md
```diff
@@ -6,8 +6,14 @@ sidebar_label: Learning Objectives
...
 
-### Data fetching (advanced)
+### Data fetching
 
+Show some data from an API in a component.
+
+- Can explain why a loading state is necessary when fetching data as the result is not available on the initial render
+- Be able to trigger `fetch()` using the `useEffect()` callback, with an empty array (`[]`) as the 2nd dependencies argument
+- Be able to store data from a successful API call in state
+- Be able to catch an unsuccessful API call, store the error in state and render an error message
```
illicitonion: We don't cover this in-class - is this expected to be pure carry-over from having done `fetch` before? Worth throwing in an exercise, maybe with an API we publish?

docs/react/week-3/learning-objectives.md
```diff
@@ -6,8 +6,14 @@ sidebar_label: Learning Objectives
...
 - Allow effects to update in response to prop changes
   - Explain why a component with an effect dependent on props is broken with empty `useEffect()` dependencies (`[]`)
   - Be able to fix a component with an effect dependent on props using the `useEffect()` dependencies
```
illicitonion, line 18: Below: This class didn't cover `onSubmit` at all, and in particular didn't cover `event.preventDefault()` - should it?

docs/react/week-3/further-reading/container-components.md
```diff
@@ -0,0 +1,17 @@
+---
+id: container-components
+title: React Further Reading - Container Components
+---
+
+In real world applications, the things we want to remember in state follow the [_business logic_](https://en.wikipedia.org/wiki/Business_logic) required by our users. So for example the number of caught Pokemon in the exercise increases when you click on the button _Catch Pokemon_. Most of the time, business logic is about figuring out when and how to change state.
```
illicitonion, line 6: This touches on one thing that I think we should emphasise more in the course:

React is all about breaking down problems into three things:
* Identifying state
* Deciding what to present given state
* Working out how/when/why to change state

I've seen many students able to follow instructions to create React components, but struggle to work out what components they should be writing or how they should be writing them, because they don't realise they should be trying to separate out these things.

In effect, our three weeks are each covering one of these (week 1 = what to present, week 2 = using state, week 3 = changing state). It feels like we can't _really_ teach this "how to design solutions" aspect until they've already seen the three separate pieces. Maybe this warrants turning React into a four-week class, where we spend week 4 talking about component design? I feel like this is too important to be a further reading footnote...

illicitonion (commented):
This is looking really great, thanks for putting it together! I think the presentation is definitely clearer, and the pacing is better too.

I realise that most of my comments here are about prose which was already there and you haven't touched (I figured the best way to review was to read the whole module top to bottom), so feel free to ignore and you want and I can make a follow-up PR with anything you don't want to take on yourself :)

Things I think are worth serious thought, though:
* We should probably be more explicit about explaining, and re-explaining, what rendering is, and how it relates to the functions people are writing.
* Students struggle with the React documentation because of how class-based it is, we should consider giving guidance text when we link to the React docs to help them read them.
* I would love to add a React week 4 where we cover "how to approach problems, break them down, and design components" :)

docs/react/week-1/lesson.md
```diff
@@ -145,16 +128,16 @@ There are 3 important parts in this code:
...
+| 2. Delete everything in the file except the line containing `export default App`. You should see an error in your terminal and in your web browser - don't panic! We're going to remake the `App` component ourselves.                    |
+| 3. Import the React variable from the React package.                                                                                                                                                                                      |
+| 4. Create a function named `App`, which will be our component.                                                                                                                                                                            |
+| 5. Within the `App` function, return a `<h1>` element with the text "Welcome to the Pokedex". What do you see in your web browser?                                                                                                        |
+| 6. Create a `<div>` element that _wraps around_ the `<h1>` you just created.                                                                                                                                                              |
```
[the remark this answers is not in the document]
alismith-dev: Right. I think this is something that often gets mentioned in the lesson, but you're right that there's no explicit coverage in the syllabus. I would be in favour of adding it.

I think the question is mostly about where it goes: week 1 is already pretty crammed.

docs/react/week-1/lesson.md
```diff
@@ -185,14 +168,14 @@ We are also using some shorter syntax within the `HelloWorld` component. `<Greet
...
+| 1. In your `pokedex` React app, open the `src/App.js` file.                                                                                                                        |
+| 2. Create a new function named `Logo`.                                                                                                                                             |
+| 3. Copy the `<header>` element and its contents and paste it into the `Logo` component.                                                                                            |
+| 4. Replace the `<header>` element in the `App` component with the new `Logo` component.                                                                                            |
+| 5. Create a new component function named `BestPokemon` and return a `<p>` element with some text saying which is your favourite Pokemon (e.g. "My favourite Pokemon is Squirtle"). |
+| 6. _Render_ your new `BestPokemon` component below the `Logo` component within the `App` component.                                                                                |
 
 #### Arrow Functions for shorter syntax
```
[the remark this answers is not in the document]
alismith-dev: Thanks that's a good spot. I think this is a remnant from before we formatted all the code snippets with Prettier. I believe it used to wrap with parens, but for a single element, Prettier will strip the parens.

So I think there's two options here:

1. We make the examples here have more than 1 element so that Prettier doesn't strip the parens
2. We remove the wording about parens, and assume that Prettier will auto-format the student's code (i.e. lack of knowledge of parens is fine because the tool will handle it for them)

I think I'm leaning towards option 1.

docs/react/week-1/lesson.md
```diff
@@ -423,31 +407,30 @@ Or calculating new values:
...
+| 9. Paste the `abilities` variable into `App.js`.                                                                                                                              |
+| 10. Pass the `abilities` variable as a prop to `BestPokemon` from `App.js`.                                                                                                   |
+| 11. In the `BestPokemon.js` file replace the existing usage of `abilities` with the `abilities` **prop**. You should still see the Pokemon ability names in your web browser. |
+| 12. **(STRETCH GOAL)** Repeat the process with the `date` variable in the `CaughtPokemon.js` file.                                                                            |
 
 ## Further reading
 
 Fed up of having to provide a prop for every component? Do you want to make your component use a value most of the time, but it can be overridden with a prop? This is a good time to use `defaultProps`. [This page on the React documentation](https://reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) describes in more detail.
```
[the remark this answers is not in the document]
alismith-dev: Thanks, good point. I don't think I reviewed this when I was looking at the initial Hooks content. It is a bit frustrating that all of the official docs haven't been updated, but my understanding is that there is an on-going effort to update them.

I did some poking around and I discovered that there's a [proposal](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md) to deprecate `defaultProps` on function components anyway (although I will note that was last year and I haven't heard any more about this since. They are supported in the latest version).

The proposal suggests moving to prop destructuring + default values, which I think is indeed too complex for it to be worth it.

So, I think I'm leaning towards removing this section altogether. EDIT: I went ahead and removed it.

docs/react/week-3/lesson.md
```diff
@@ -134,7 +272,7 @@ Together let's "play computer" to break down exactly what is happening with thes
   }
 ```
 
-8. Now `App` **does** render `MartianImageFetcher` and passes the `date` state as a prop (also named `date`)
+8. Now `App` **does** render `MartianImageFetcher` and passes the `date` state as a prop named `photoDate`
 9. The `MartianImageFetcher` function component is called for the first time
 10. `useState` knows that we haven't got any `imgSrc` state so initialises it to `null`
 11. We queue an effect, which will run after we render for the first time
```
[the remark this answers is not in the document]
alismith-dev: Thanks, good catch. Originally it was mentioned in the conditional rendering section, but I guess I removed it when I was streamlining that section. I'll see if I can find a good place for it.

docs/react/week-3/lesson.md
alismith-dev: Thanks, this is good feedback.

I'm a bit in two minds here: I was intentionally avoiding the deps array in the "The timeline of this component" para, because I wanted to get across the "loop" concept of setting state inside the fetch first. I didn't want to get into too much detail that distracts from this point. The "walkthrough" section was designed to introduce the deps array separately.

However, I do agree that it makes point 18 too confusing and out-of-context. I therefore added a sentence below the "The timeline of this component" para which calls out that the effect wasn't run a second time, but says that it'll be covered later.

I also reworded point 18 a bit so that it doesn't explicitly call it the "dependencies array". Feedback welcome

docs/react/week-3/lesson.md
alismith-dev: > Very frequently they try to write;

That is why we have such an extensive event handling section in week 2 🙂 (IME most React tutorials hardly touch on this).

But I do agree that it's one of the most common problems for our students.

> single event handler function is a bit involved and fiddly

I think the feeling was that this is a common enough pattern that it would be worth showing.

> little arrow functions are concise

We've gone back and forth on multiple different implementations in this section. I am still unsure what the best option is.

In the past, my hesitation with inline event handlers is that they are somewhat tricky to debug:

```js
<input onClick={event => console.log('foo') setPassword(event.target.value)} />
```

I also think it leads to a bad habit where students dump all the logic inline, leading to poor readability.

I however could see an argument for showing this implementation with separate function definitions first, and mentioning that they could be inlined.

> Maybe this lesson is already long enough

Yes indeed, my perception is that this lesson will already be quite challenging to fit into one day.

My proposal:

1. We switch the first implementation to multiple separate function definitions
2. We then mention that they could be inlined, as with your example
3. Finally, we show a centralised implementation (i.e. the existing one), and mention validation

I think this matches the order of importance of the concepts:

1. Handling multiple form fields
2. Practicing event handling
3. Writing form code concisely
4. Showing patterns for centralised code

Which allows the teacher to cut off if out of time.

Thoughts?

docs/react/week-3/lesson.md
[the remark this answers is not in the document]
alismith-dev: I ended up adding a note to the "The timeline of this component" bit.

docs/react/week-3/lesson.md
```diff
@@ -246,13 +384,18 @@ Phew! That was a lot of work just to render an image! But we're not quite done y
...
-| **Exercise B**                                                                                        |
+| Exercise D (estimate: 5 min)                                                                          |
 | :---------------------------------------------------------------------------------------------------- |
 | 1. Did you spot where the bug was? Discuss with a group of 2 - 3 students where you think the bug is. |
 | 2. Report back to the rest of the class where you think the bug happened.                             |
 
+<details>
+<summary>Click here to reveal the answer!</summary>
+
 The key that the `useEffect` in `MartianImageFetcher` is **only run once**. This is because we told React that the queue should be queued on the first render only. However, as we saw, sometimes you need the effect to run again when some data changes. In this case the `date` prop, changed from `"2019-01-01"` to `"2020-01-01"`, meaning that we have to fetch data different data. We call this a _dependency_ of the effect.
```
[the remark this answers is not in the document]
alismith-dev: I reworded this suggestion a bit, but thanks!

docs/react/week-3/learning-objectives.md
alismith-dev: > is this expected to be pure carry-over from having done fetch before?

Hmm, good point. I think the concept of an error state would be new.

I am a bit concerned about time in week 3 though.

docs/react/week-3/learning-objectives.md
alismith-dev: Yep it probably should, but again concerned about the time.

docs/react/week-3/further-reading/container-components.md
alismith-dev: This is a great suggestion, but as I think you're aware, we don't really have time to cover this in any depth.

> Maybe this warrants turning React into a four-week class, where we spend week 4 talking about component design?

This was an idea that came out of the London post-module retro too.

I would welcome proposals/PRs with more specifics! 😉 

---

Apologies, the rest of this comment isn't really about React. Perhaps we should raise it on Slack/in another issue for wider visibility

I think structural changes like this would have to be discussed with the wider group. The 3 week constraint is more of a limit on volunteer time than anything else, so we'd have be careful with it (speaking from experience, I like to take some time off after the 3 weeks). I think the organisation in general would have concerns about further lengthening the course.

Additionally there's a couple of thoughts I have about it:

- I would feel a bit weird about having 4 weeks of React, but only 3 weeks of Node & DBs
- Not quite sure if putting it directly after the 3 existing weeks is the best choice
- Perhaps a "best practices" module at the end of the course might work better? This may also allow making it optional for students that are struggling?

40thieves:
Thanks so much for this detailed review @illicitonion! It's fantastic to get such good feedback 🙂 

> We should probably be more explicit about explaining, and re-explaining, what rendering is, and how it relates to the functions people are writing.

Could you expand on this a bit, please? I'm not exactly clear on what aspects you mean.

I have re-worked the "Re-rendering components" section multiple times. My intention was to try isolate the concept of re-rendering (and to some extent the virtual DOM) from state. But I've found it difficult to do without getting into the implementation details of how to trigger a re-render.

In the state section itself, I really tried to emphasise the aspect of triggering a re-render via state changes. If it's not getting through, I'd appreciate a second perspective.

> Students struggle with the React documentation because of how class-based it is, we should consider giving guidance text when we link to the React docs to help them read them.

I addressed one area that you highlighted (by removing it entirely). If there's other specific areas in the syllabus I can take a look at those too.

Unfortunately I don't know there's much we can do about students finding materials on their own time. Do you think we should explicitly discuss (in the syllabus) that class components exist? I believe we did talk about it during the class, but I don't think it's written down anywhere.

My understanding is that there is an effort to significantly re-work the official docs to use more function components. So we may just be able to get away with waiting on that (given that the next class won't be until next year).

> I would love to add a React week 4 where we cover "how to approach problems, break them down, and design components" :)

I think I've addressed that in https://github.com/CodeYourFuture/syllabus/pull/36#discussion_r467458270.

docs/react/week-1/lesson.md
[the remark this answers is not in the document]
illicitonion: Yeah, #1 feels like it's going to be simpler, and reinforces to the students that parens may be required (and if they put more parens than they need, that's not the end of anyone's world)

docs/react/week-3/lesson.md
illicitonion: I think this looks great, a huge improvement!

One thing we may want to do is also add a `console.log("Fetching data from NASA");` line in the `useEffect` callback before the `fetch`, so that we can more obviously demonstrate that the callback isn't being called again. As things currently stand, we say "You might notice that even though we re-rendered, we did not run the useEffect a second time", but we haven't really _shown_ the students that the callback isn't called again (for all they know it is called again, and is returning the same result so not re-rendering, or even _is_ re-rendering just very quickly, or something) - being able to say "The long only happened once, so the function was only called once" could be a nice demonstration.

Another brief thought: I _think_ we've ended up in a place where we don't actually define the word "dependency" any more... We have the text:
```
Because of this behaviour, the second argument to useEffect is called the dependencies argument. We use it whenever we have something in our effect function that depends on a variable outside of the effect function.
```
But I think the first of those sentences pre-supposes being familiar with the concept of a dependency... Maybe we could add a first sentence to the "`useEffect` dependencies array" section, something like: "Our callback function uses the value of the `photoDate` `prop` - what the function does will depend on what value of `photoData` was used, so we call `photoData` a _dependency_ of the calllback function."?

docs/react/week-3/lesson.md
illicitonion: Sounds great! :)

docs/react/week-3/learning-objectives.md
illicitonion: I'd suggest just dropping the objective. We kind of hand-wave over `fetch` errors in JS2 - if we're going to try to teach handling them better in React, we should start by bulking up their treatment in JS2.

docs/react/week-3/learning-objectives.md
illicitonion: This _is_ one I've seen people struggle with in exercises, particularly when they're working on the Chat App later on... Do you think we could squeeze it into week 2 as a 10 minute exercise which follows on from click handlers? Maybe a "Give my caught pokemon a nickname" form? We can hand-wave over `event.preventDefault()` by just saying something like "We need to do this whenever we handle a form in React, if you're interested in why, [here](https://www.robinwieruch.de/react-preventdefault)'s some further reading"

docs/react/week-3/further-reading/container-components.md
illicitonion: Definitely agree this is a discussion for another venue - wrote up https://codeyourfuture.slack.com/archives/C012UUW69S8/p1596893069021300 :)

illicitonion:
I think you've covered all of these perfectly in responding to the line-comments :) Sorry, this PR-comment was meant to say "You can ignore most of my line comments if you want, but it would be good to make sure to address the ones pertaining to these topics" - by acting on all of them, you made this comment unnecessary :D

> > We should probably be more explicit about explaining, and re-explaining, what rendering is, and how it relates to the functions people are writing.
> 
> Could you expand on this a bit, please? I'm not exactly clear on what aspects you mean.
> 
> I have re-worked the "Re-rendering components" section multiple times. My intention was to try isolate the concept of re-rendering (and to some extent the virtual DOM) from state. But I've found it difficult to do without getting into the implementation details of how to trigger a re-render.
> 
> In the state section itself, I really tried to emphasise the aspect of triggering a re-render via state changes. If it's not getting through, I'd appreciate a second perspective.

I think you've got this pretty much perfect now :)

I think the one additional suggestion I'd have is to take the week 1 sentence "We render the HelloWorld component into the element with the id of root." and add a phrase like " - The JSX elements that are returned from the `HelloWorld` function are turned into the DOM for you by React, and put inside the root element", like you emphasised in the re-rendering components section in week 2.

> > Students struggle with the React documentation because of how class-based it is, we should consider giving guidance text when we link to the React docs to help them read them.
> 
> I addressed one area that you highlighted (by removing it entirely). If there's other specific areas in the syllabus I can take a look at those too.

That was what I was mostly aiming for :)
 
> Unfortunately I don't know there's much we can do about students finding materials on their own time. Do you think we should explicitly discuss (in the syllabus) that class components exist? I believe we did talk about it during the class, but I don't think it's written down anywhere.

I think it's probably worth mentioning in the syllabus. If we're viewing the syllabus as a way for different cities to pick up the same material and deliver it, anything not explicitly mentioned is liable to be left out for some classes :) Maybe it could be a "further reading" section at the end of week 1, so as to not disrupt the flow of the lesson? Gives us something easy to point to on Slack if students are confused, too.

illicitonion (approved):
This is looking wonderful, thanks so much for taking the time to act on everything!

40thieves:
> I think the one additional suggestion I'd have is to take the week 1 sentence "We render the HelloWorld component into the element with the id of root." and add a phrase like " - The JSX elements that are returned from the HelloWorld function are turned into the DOM for you by React, and put inside the root element", like you emphasised in the re-rendering components section in week 2.

I added a bit more to the definition of rendering, to match the re-rendering wording in week 2. Thanks that is a good suggestion.

40thieves:
Just pushed another update to add the proposed form handling changes (including a submit handler).

docs/react/week-3/lesson.md
alismith-dev: EDIT: I added the log, that's a good idea.

> I think we've ended up in a place where we don't actually define the word "dependency" any more...

We have the hidden answer to Exercise D, which has this text:

"We call this a _dependency_ of the effect. Any variables which are used inside the `useEffect` callback function are dependencies."

I'm unsure if this is obvious enough that everyone would find it though?

I originally hid it because it would be trivial to answer Exercise D by just scrolling down a bit. I could just un-hide it and assume that we could have discussion anyway?

40thieves:
I'm undecided on adding an exercise for form submission. And to be honest, I'm kind of done with looking at text editors for today. Hopefully everything else has been addressed.

illicitonion (approved):
Looks great, thanks again so much!!

PR #41, opened 2020-08-02
Always use Yarn

textbook:
The `package-lock.json` file was out-of-date, so `npm ci` bailed anyway.


40thieves:
Hmm, could we consider the other way around? I don't have Yarn installed and given that NPM is bundled with Node, I suspect that's the case for a majority of folks.

I wouldn't be surprised to see someone re-introducing a `package-lock.json` by accident in the future. At least if you're using Yarn, you'd know to ignore the `yarn.lock` when committing.

Or alternatively, we should document the usage of Yarn over NPM here.

textbook:
@40thieves Yarn is the one that worked, and the workflow gives it precedence. I'd be just as happy (maybe happier, even) to switch it over to NPM, but I went with what was currently working.

> I wouldn't be surprised to see someone re-introducing a `package-lock.json` by accident in the future.

I added a little something-something to make sure that doesn't happen...

ChrisOwen101:
Yarn is the "encouraged" way to manage Docusaurus according to their documentation

https://v2.docusaurus.io/docs/installation

DomVinyard:
In my purely anecdotal experience, npm is again becoming the favoured package manager

textbook:
I'd be happy to supersede this with a switch to just use NPM, but at the moment we have both `package-lock.json` (currently outdated) _and_ `yarn.lock`, which I think is a bit confusing.

40thieves:
Agree that it is confusing and we should pick one 👍 I really don't mind using Yarn if that's the choice going forward.

I did notice recently that it is documented in [the contributing guide](https://syllabus.codeyourfuture.io/contributing/technical), so that covers my primary concern that folks would default to NPM.

I do agree with Dom that NPM feels like it's the more common choice now. Just in case we want to go with that option, I worked up: https://github.com/CodeYourFuture/syllabus/compare/remove-yarn-lock Which I believe is the equivalent change, but using NPM instead (I think I got all the packages up-to-date too). I took a slightly simpler approach to blocking `yarn.lock`, which still allows people to use Yarn, they just can't check in the lockfile.

> Yarn is the "encouraged" way to manage Docusaurus according to their documentation

Suspect there's a bit of Facebook favouritism going on there. I'm not aware of any particular reason that Docusaurus wouldn't work with NPM.

textbook:
@40thieves nice! I think the problem with just ignoring `yarn.lock` is that the install itself can leave `package{,-lock}.json` out-of-sync again, so `npm ci` will fail. We could ignore all of the lock files and just `npm install`, but then our builds aren't so reproducible. For forcing not-Yarn there's a slight hack that's even easier than a package script: https://stackoverflow.com/a/58087572/3001761.

40thieves:
>  I think the problem with just ignoring yarn.lock is that the install itself can leave package{,-lock}.json out-of-sync again

My assumption was that Yarn users would be sufficiently advanced to realise that their `yarn.lock` was ignored and thus would work around it. But yeah, perhaps that is a bit of a leap.

textbook:
Superseded by #59

PR #43, opened 2020-08-03
Fix typo in asset name

textbook:
For building on case-sensitive file systems like in the GitHub Action...


textbook:
Looks like you fixed it simultaneously @ChrisOwen101 😭

ChrisOwen101:
Ah! Sorry @textbook!

PR #56, opened 2020-08-19
[SQL] Updates from Manchester

ChrisOwen101:
## What does this change?

Module: SQL
Week(s): 1/2/3

## Description

Keith Bremmer (of Manchester) has put in some time to improve the SQL module. 

Lots of varied improvements in there and lots of new exercises.


llh1 (changes requested):
I've started commenting on the PR but then I realised the scale of the proposed changes, so I gave up commenting on the specifics for now. Probably easier to comment on the general PR here then we can discuss the specifics later :) I don't understand the purpose of these changes, it looks more like a complete rewrite of the current DB syllabus instead of building upon and improving the existing. 

Overall, sorry... but I think I disagree with quite a lot of the proposed changes. My main concern is that it's way too complex for the students as it goes into too many granular details which are not necessary in an introduction to DB and SQL.

I'm also struggling to understand the progression of the class, querying data with SELECT is introduced before talking about tables which seems confusing to me. Also not sure why the DB modelling exercise is removed? It's something which worked really well in Glasgow. To me it was critical for a good understanding of how to design a DB while introducing the important concepts of tables, columns, rows and relationship between tables etc... without talking about SQL yet.

For the Install Guide, i personally wouldn't put it in the syllabus as i don't think it's relevant to what we are teaching. In Glasgow for example, we provision all the machines with Ubuntu and postgresql installed by default.

I think it would help to understand what are the problems we are trying to solve with this PR and what doesn't work in the current syllabus first. Then we can address these problems and improve the syllabus in several other PRs, ideally one for each week once we have an action plan.

PR #59, opened 2020-09-01
Always use NPM

ChrisOwen101:
This is the opposite version of this PR

https://github.com/CodeYourFuture/syllabus/pull/41

It removes the use of Yarn and leaves us only using NPM


textbook:
I looked at resolving the conflicts, but they seem to be due to most of the changes having already been made. I think post #61 there's nothing left to merge from this, so we can probably close it out.

PR #61, opened 2020-09-06
Strip out conditional installation

textbook:
## What does this change?

Module: N/A
Week(s): N/A

## Description

There should no longer be a `yarn.lock`, and the build should fail if the NPM lockfile is out-of-sync. Therefore to simplify the build process, we can strip out the conditional installation.

@ChrisOwen101 - I think this is the only part of #59 that hasn't already made it into `master`!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,13 +17,7 @@ jobs:
           node-version: "12.x"
       - name: Test Build
         run: |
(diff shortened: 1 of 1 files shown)


ChrisOwen101:
Thanks @textbook !

ChrisOwen101:
@all-contributors add @textbook for infra

PR #78, opened 2020-10-06
Fix Button Border Color on Homepage

### LINKED ISSUE #62: Button styling not working on Production build
**Describe the bug**

On the production build of the Syllabus website the button does not have it's rounded button styling

See here for production:
image

It should look like how it does on a local build. See here:
image

**What should happen**

The button should be styled the same on Production and Local Builds

ChrisOwen101:
A fix that sets all border colours to be the same as the button color
Closes #62


illicitonion (approved):
Perfect! 💯

PR #87, opened 2020-10-15
What makes a good final project?

### LINKED ISSUE #83: What Makes A Good Final Project
Module(s): Final Project
Week(s): N/A

When our students reach Final Projects, we like it if they can work on projects for NGOs/Charities. 

Some of our City Coordinators are asking what we should look for in a project to make sure that our students have a good experience. 

Making a list of previous external projects would be nice.

You can see a list of internal projects here:
https://syllabus.codeyourfuture.io/finalproject/projects

These projects all fulfil all the criteria of a good Final Project.

Making sure this task is well-defined will reduce the risk of an unsuitable project being worked on by our students.

**Additional context**

Chat with Rob Crook (Product Manager from Spotify) for thoughts. He's been running Final Projects for us in two cities.
Claire Blickley & Barnaby Dathan are City Coordinator for West Midlands and North West respectively.

ChrisOwen101:
Closes #83

### DIFF
--- a/docs/finalproject/external-projects.md
+++ b/docs/finalproject/external-projects.md
@@ -4,7 +4,7 @@ title: External Projects
 sidebar_label: External Projects
 ---
 
-For Final Projects we try to find interesting, tech-for-good projects for that stretch the abilities of those working on them whilst helping an NGO or Charity work towards its goals.
+At CodeYourFuture we try to find interesting, tech-for-good projects for that stretch the abilities of our students whilst helping an NGO or Charity solve a business problem they are facing.
 
 ## Past Projects
 
--- a/docs/finalproject/external-projects.md
+++ b/docs/finalproject/external-projects.md
@@ -18,6 +18,30 @@ Additionally, here are some examples of internal projects we've completed
 
 You can also find more example project briefs on the Project Brief
 
+## Goals
+
+With the final projects, we’re looking to give our students:
+
+- Experience of collaborating in a team to deliver software
+- An opportunity to work with real stakeholders and requirements
+- The chance to bring together the different things they’ve learned
+- A real example to talk about in job interviews and guide their interests
+
+and to give our partners:
+
+- A useful piece of software that they can use as-is or build on after the project
+
+More details can be found on the [Intro](./intro) and [Preparation](./prep) pages.
+
+## What does a Product Owner have to do?
+
+To be a good Stakeholder/Product Owner you should
+
+- Be able to attend a weekly/bi-weekly meeting with our teams
+- Be in a position to advise the development of the project
+  - e.g. Should we prioritise Feature A or Feature B?
+- Have realistic expectations of what our students can build in the time given to them
+
 ## What is a Good Project?
 
 It's important that the projects that we work on fulfill some base requirements to make sure they go smoothly.
@@ -26,7 +50,7 @@ All Final Projects aim to be a [Minimum Viable Product](https://www.agileallianc
 
 ### Functionality
 
-A Good Final Project should have most of the following qualities
+A Good Final Project should have the following qualities
 
 1. Be an interactive website
 
@@ -60,6 +84,13 @@ We use the following technologies at CodeYourFuture
 An external Final Project should use all of these technologies into a single Full Stack package. This is important as we want our students to use all of the skills that they have learnt in our course.
 
 We advise against using other tools as they will fall outside our students areas of competency due to the time constraints that we have.
+
+### Limitations
+
+As Junior Developers, our students do have some limits on what they can produce
+
+- Visual Design Complexity
+  - An overly complex design will take a lot longer for our students to develop. Try to keep your designs simple and functional.
 
 ## Next Steps
 
--- a/docs/finalproject/external-projects.md
+++ b/docs/finalproject/external-projects.md
@@ -23,7 +23,7 @@ You can also find more example project briefs on the Project Brief
 With the final projects, we’re looking to give our students:
 
 - Experience of collaborating in a team to deliver software
-- An opportunity to work with real stakeholders and requirements
+- An opportunity to solve real world business problems
 - The chance to bring together the different things they’ve learned
 - A real example to talk about in job interviews and guide their interests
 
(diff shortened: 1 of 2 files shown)


ChrisOwen101:
Feedback from Martin Bate

- Where you mention 'real stakeholders and requirements', in my experience it can sometimes help to flip the wording onto 'problems' rather than requirements to avoid the client (in this case the Charity) becoming too prescriptive
- It might be an idea to work with the client to generate the designs - again they may not have a feel for what is easy/difficult and may have unreasonable expectations of what is achievable based on the apps/sites they sue everyday.
- I would clarify that it's really helpful if the PO can be available for messages etc all the time, and as a minimum once or twice a day. Students may likely become blocked, spin wheels or go off on a tangent if they have a burning question and can't get it answered. It might be worth considering that the PO as you've described them here is really a more traditional 'client' and that you have a mentor or similair fulfil the PO role as a proxy for the client (so that they can be more available for the students).

docs/finalproject/external-projects.md
```diff
@@ -0,0 +1,127 @@
...
+
+More details can be found on the [Intro](./intro) and [Preparation](./prep) pages.
+
+## What does a Product Owner have to do?
+
+To be a good Stakeholder/Product Owner you should
+
+- Be able to attend a weekly/bi-weekly meeting with our teams
+- Be in a position to advise the development of the project
+  - e.g. Should we prioritise Feature A or Feature B?
```
alismith-dev, line 42: Perhaps this covers it already, but maybe something like "Be willing to cut scope" (or an equivalent non-jargon phrase)?

docs/finalproject/external-projects.md
```diff
@@ -0,0 +1,127 @@
...
+All Final Projects aim to be a [Minimum Viable Product](https://www.agilealliance.org/glossary/mvp/). This means they may be rough around the edges but have all of the core functionality working.
+
+### Functionality
+
+A Good Final Project should have the following qualities
+
+1. Be an interactive website
+
+   - This means a website that a user can interact with to create, read, upload, delete data.
+   - For example, a photo hosting service or blog would fulfil this criteria
```
alismith-dev: I'm a bit hesitant about the blog example. I would guess that for most third-party stakeholders Wordpress and friends would be much more suited to their needs. (Which we obviously wouldn't be able to do).

Some ideas for alternatives: custom events or Q&A service.

If we're going with a blog example, then maybe we could frame it as "custom blog functionality" or something?

docs/finalproject/external-projects.md
```diff
@@ -0,0 +1,127 @@
...
+We use the following technologies at CodeYourFuture
+
+- JavaScript
+- React
+- Node with Express
+- MongoDB or PostgreSQL
+
+An external Final Project should use all of these technologies into a single Full Stack package. This is important as we want our students to use all of the skills that they have learnt in our course.
+
+We advise against using other tools as they will fall outside our students areas of competency due to the time constraints that we have.
```
alismith-dev: I would make this stronger, as I think it's pretty much a hard constraint?

docs/finalproject/external-projects.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Limitations
+
+As Junior Developers, our students do have some limits on what they can produce
+
+- Visual Design Complexity
+  - An overly complex design will take a lot longer for our students to develop. Try to keep your designs simple and functional.
+- Instant Answers
+  - You shouldn't expect our students to be able to give instant answers to any technical questions you might have. They will get there with time however!
```
alismith-dev, line 100: Not sure if this is the best place to mention this, but do we want to touch on volunteer support? Maybe worth covering off the idea that volunteers are there to help the students not necessarily the project itself?

docs/finalproject/external-projects.md
```diff
@@ -0,0 +1,127 @@
...
+
+A good next step to getting started with defining your project is to complete this form.
+
+[CodeYourFuture - Projects Specification Form](https://docs.google.com/forms/u/1/d/1u2zImwEjD5PDWJuMyDeCuQ3AP5xhWr683pGy9HlmC14/edit)
+
+It is important that you answer all of these questions in **as much** details as possible. Even irrelevant details will help us understand what you are trying to achieve.
+
+### 2) Visual Form
+
+On the new step we're asking you to be more visual in answering the questions. We'd advise you to use images or diagrams to answer the questions in these slides.
```
alismith-dev: I think this is supposed to be:

```suggestion
On the next step we're asking you to be more visual in answering the questions. We'd advise you to use images or diagrams to answer the questions in these slides.
```

docs/finalproject/external-projects.md
```diff
@@ -0,0 +1,127 @@
...
+
+[CodeYourFuture - Visual Specification Form](https://docs.google.com/presentation/d/1-rj0NAncIhIZuay6TXEyUeuXFFUnE7vmN12223rHZOI/edit#slide=id.g3579738be0_0_0)
+
+### 3) Design Mock-Ups
+
+Finally, you should work with us to create some simple designs of your product. These will not be the final designs of what you be building - they're just a tool to help us work out what you want built.
+
+Very quick designs of the tool/website you are trying to build will give us a great insight into what exactly you want built.
+
+You needn't use any specialised tools to design this. Paint or Pen & Paper will do for this step.
```
alismith-dev: Would linking to a basic wireframing tool be helpful? I was going to suggest Balsamiq, but looks like it's a free trial only now. Maybe https://excalidraw.com/?

docs/finalproject/external-projects.md
```diff
@@ -0,0 +1,127 @@
...
+
+Try to think about how each individual page will look. Will the user need to create an account? Think of the form they will need to fill in. Does the user view content on a page? What content should be included?
+
+Get in contact with your local City Coordinator to arrange this.
```
alismith-dev, line 128: Do we have an contact page or something we could link to?

alismith-dev (approved):
A few minor points, but nothing blocking :shipit:

docs/finalproject/external-projects.md
ChrisOwen101: By this point they should have contact with their City Coordinator - I imagine this page will be sent to the NGOs directly rather than them finding it by themselves

PR #93, opened 2020-10-16
Add Task Numbers to Class Exercises for Node

ChrisOwen101:
## What does this change?

Module: Node
Week(s): All

## Description

Added Task Numbers to make it easier to understand the flow of the lesson. 

For example

Before:
image

After:
image

### DIFF
--- a/docs/node/week-1/lesson.md
+++ b/docs/node/week-1/lesson.md
@@ -68,7 +68,7 @@ What's an API and how is it different from a web site?
 
 ## 2) Intro to Node on Glitch
 
-### 2.2) Running a Simple Server
+### 2.1) Running a Simple Server
 
 Run the simplest web server code.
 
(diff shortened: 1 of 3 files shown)


IsmaelMartinez:
Hi @ChrisOwen101 , will it not be possible to use the markdown ordered list on this? Using a `.` instead of a `)`. Then you can create a simple Table of Content if required/suitable

ChrisOwen101:
The table of content is actually generated by Docusaurus so we don't need to worry about that so much :)

See [here](https://syllabus.codeyourfuture.io/node/week-1/lesson).

IsmaelMartinez:
I see what you referring. The headers... and markdown doesn't support numbering in headers. So ignore my previous suggestion :) Let me know if you want me to go throw some of the pages... you got a few to modify!

alismith-dev (approved):
Happy to go with this as-is :shipit: 

But I wonder if the numbering is needed at all? I don't believe we use numbering in any other modules? Especially now we that we have the persistent ToC, the ordering I think is implicit?

PR #94, opened 2020-10-16
Add Node "Presentation" Homework

ChrisOwen101:
## What does this change?

Module: Node
Week(s): 1/2/3

## Description

In London last year we ran an exercise where student had to

1. As Homework, create a short presentation about what they learnt in the lesson
2. Present it to small groups of people 

This is cheap, quick and encourages students to engage more with the content. Presentation practice is also valuable.

It went well in London so I thought it'd be good to role this out to other cities.

### DIFF
--- a/docs/node/week-1/homework.md
+++ b/docs/node/week-1/homework.md
@@ -4,10 +4,24 @@ title: Homework
 sidebar_label: Homework
 ---
 
+## Coursework Repo
+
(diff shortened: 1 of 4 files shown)


ChrisOwen101:
Apologies, had to fast track this through before a lesson on Sunday so had to skip the approval step

PR #107, opened 2020-10-26
Add More React Preparation Work

ChrisOwen101:
## What does this change?

Module: React
Week(s): Preparation

## Description

This short exercise was given to London Class 6 and seemed to have a good results - so I'm moving it into the Syllabus.

I believe this was created by @nbogie 

@40thieves

### DIFF
--- a/docs/react/prep.md
+++ b/docs/react/prep.md
@@ -6,7 +6,7 @@ sidebar_label: Preparation
 
 Before you start React you must make sure that you have some additional software installed.
 
-## Node & NPM
+## 1) Node & NPM
 
 You should already have these tools installed.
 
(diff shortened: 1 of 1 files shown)


docs/react/prep.md
```diff
@@ -27,3 +27,18 @@ Please follow this setup guide to get a new piece of software installed.
...
+   - You can stop watching at 11:06 (the rest is unnecessary)
+
+2. Answer these questions about the video
+
+   - A. What does he say the "best description of React is"?
+   - B. What does he say "could be a component"?
+   - C. What does he say is the benefit of using JSX?
+   - D. What makes React fast?
+
+3. Once you have answered the question, submit them on Slack or hand them in online.
```
illicitonion: Can we suggest one thing to do rather than two?

illicitonion (approved):

PR #109, opened 2020-10-26
Add MongoDB Prep Work

ChrisOwen101:
## What does this change?

Module: MongoDB
Week(s): Preparation, Week 1

## Description

A small piece of setup is required before teaching MongoDB.

All student should

1. Sign Up to MongoDB Atlas
1. Install MongoDB Compass

This task was taken from London Class 6's Google Classroom

@mickyginger

### DIFF
--- a/docs/mongodb/index.md
+++ b/docs/mongodb/index.md
@@ -4,6 +4,14 @@ title: MongoDB
 sidebar_label: Overview
 ---
 
+## Preparation
+
+Before starting this module there is some [preparation work](./preparation) to get you setup.
+
+Make sure you complete it before attending class.
+
+## Content
+
 | Week                         | Content        |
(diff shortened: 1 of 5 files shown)


ChrisOwen101:
From @nbogie 

Looks good!  It'd be possible to offload more stuff from the first class - the creation of their first (and only) cluster, import of the sample data, and the creation of a database user, for example.
I expect we could find suitable existing documentation in a getting started mongodb & atlas tutorial - there is nothing at all custom about what we want them to do.  Or we find/record a little video.  (This may be beyond the scope of this ticket.)
With that we'd still have to provide some troubleshooting either pre-class or (worse) in the first 30 minutes of the class.
It's not ideal to have students run through a lot of set up before they know what the service is even for, but all told, it's under ten minutes of clicking links and downloading.

PR #115, opened 2020-10-28
Add Example Image of Bad Document Naming

ChrisOwen101:
## What does this change?

Module: Git
Week(s): 1

## Description

Working on the slides for this week and I added an image that I thought would be useful here as well. 

This gives a visual example of the bad naming problem that Git tried to solve.

@40thieves @illicitonion


illicitonion (approved):
Thanks!

PR #119, opened 2020-10-30
Add Marking Guide for Coursework

ChrisOwen101:
## What does this change?

Module: Guides

## Description

I'd like this to be a general, all purpose marking guide for all of our coursework at CYF and is a first step to having a rubric for every week of the homework. 

I'd also like this to be used by Students to see what they should be aiming for. For example, we can tell students to consider the Edge Cases and then refer them to this page for more details.

Very open to more areas that we should be evaluating our students on 

## Rendered Version

 [Rendered version](https://github.com/CodeYourFuture/syllabus/blob/marking-guide/docs/guides/marking-guide.md)

### DIFF
--- a/docs/guides/marking-guide.md
+++ b/docs/guides/marking-guide.md
@@ -199,7 +199,7 @@ calculateScreenSize(1800);
 
 Will be hard to understand when you codebase grows.
 
-Instead us constants to store your hard coded values. For example
+Instead us constants to store your hard coded values - this means the value has a name, which can be [usefully descriptive](#good-variable-names). For example
 
 ```js
 let ENDPOINT = "https://www.api.com/retrieve";
--- a/docs/guides/marking-guide.md
+++ b/docs/guides/marking-guide.md
@@ -39,7 +39,11 @@ This Rubric should be used to guide your assessment of Students Homework at Code
 
 Edge cases are when code does something you're not expecting and you didn't account for in your code.
 
-For example, using a number in a `Name` field of a form might cause problems since peoples names can't be numbers.
+Some edge cases are valid data, and some are not.
+
+Some examples of valid edge case are that some names contain punctuation, or some people may only have a single name, which your code needs to be able to handle. An example of an invalid edge case is if someone doesn't enter a name at all, which your code should reject - if your code doesn't notice the problem where the user provides their name, other code you wrote may make assumptions which are incorrect.
+
+See [Falsehoods Programmers Believe About Names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) for some interesting examples.
 
 Good code should handle all of these cases so that they can't happen - and if they do - they don't make your program crash or stop work.
 
--- a/docs/guides/marking-guide.md
+++ b/docs/guides/marking-guide.md
@@ -55,7 +55,11 @@ Validations are used wherever necessary. The code never breaks under any circums
 
 ### Small Functions
 
-Writing small functions makes your code easier to understand and easier to work with
+Writing small functions makes your code easier to understand and easier to work with. Ideally you should be able to describe what a function does in one sentence, without the word "and".
+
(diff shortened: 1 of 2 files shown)


docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+sidebar_label: Marking Guide
+---
+
+This Rubric should be used to guide your assessment of Students Homework at CodeYourFuture
+
+## Overview
+
+### Completeness (/10)
+
+- The code achieves all of the tasks in the coursework
```
illicitonion: May be worth calling out something about it doing so in the intended way. Sometimes we see learners e.g. changing test cases, or over-fitting their solutions to the test cases in the homework - they should aim to identify and solve the general problem set by the task.

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,210 @@
...
+means that the variable ‘pocketMoney’ is being used to store how much pocket money you have. Right now you have £20.
+
+The name given to each variable is up to the programmer, but ideally a variable name should have meaning, ie it should reflect the value that it is holding.
+
+#### More Reading
+
+- [Good Variable Names](https://spin.atomicobject.com/2017/11/01/good-variable-names/)
+- [Programming Basics - Naming Variables](https://www.bbc.co.uk/bitesize/guides/zwmbgk7/revision/3)
+
+### More Optimal Code
```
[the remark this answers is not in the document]
ChrisOwen101: That's exactly what I was going for - thanks!

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,293 @@
...
+
+Edge cases are when code does something you're not expecting and you didn't account for in your code.
+
+Some edge cases are valid data, and some are not.
+
+Some examples of valid edge case are that some names contain punctuation, or some people may only have a single name, which your code needs to be able to handle. An example of an invalid edge case is if someone doesn't enter a name at all, which your code should reject - if your code doesn't notice the problem where the user provides their name, other code you wrote may make assumptions which are incorrect.
+
+See [Falsehoods Programmers Believe About Names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) for some interesting examples.
+
+Good code should handle all of these cases so that they can't happen - and if they do - they don't make your program crash or stop work.
```
illicitonion:
```suggestion
Good code should handle all of these cases so that they can't happen - and if they do - they don't make your program crash or stop working.
```

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,293 @@
...
+Hard coded values with no explanation make code hard to understand and read in the long term. For example
+
+```js
+getData("https://www.api.com/retrieve");
+calculateScreenSize(1800);
+```
+
+Will be hard to understand when you codebase grows.
+
+Instead us constants to store your hard coded values - this means the value has a name, which can be [usefully descriptive](#good-variable-names). For example
```
illicitonion:
```suggestion
Instead use constants to store your hard coded values - this means the value has a name, which can be [usefully descriptive](#good-variable-names). For example
```

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,293 @@
...
+
+```js
+let ENDPOINT = "https://www.api.com/retrieve";
```
illicitonion:
```suggestion
let WEATHER_ENDPOINT = "https://www.api.com/retrieve";
```

Let's be even less generic :)

docs/guides/marking-guide.md
```diff
@@ -0,0 +1,293 @@
...
+let WIDTH = 1800;
```
illicitonion:
```suggestion
let MAX_SCREEN_WIDTH = 1800;
```

illicitonion (approved):
Looks great! :)

PR #155, opened 2020-12-12
Fix confusing exercise instruction in js-core-1/week2 lesson

szemate:
## What does this change?

Module: js-core-1
Week(s): 2

## Description

The exercise asks students to create a function so that it takes an array and

>  if the second index in the array contains the name "Amy", it returns `"Second index matched!"`

however it's not the _index_ that should contain the name "Amy", but the _element_ that is pointed to by the index.

 [Rendered version](https://github.com/szemate/syllabus/blob/js-core-1/week-2/fix-exercise-wording/docs/js-core-1/week-2/lesson.md#exercise-5-mins-5)

?


illicitonion (approved):
Thanks!

PR #156, opened 2020-12-22
Add code style guide

### LINKED ISSUE #105: Code Style Guide for CodeYourFuture

In the wild, there are lots of good Code Style guides that can help us to make sure our student code in professional way. 

The problem with many corporate-grade guides is that they are very dense and assume an advanced knowledge of JavaScript - not ideal for our students when they're starting out.

For inspiration here is [AirBnBs](https://github.com/airbnb/javascript). 

We often tell our students to "write good code" but our definition of that is sparse and disparate. This file will become our single source of truth for what we expect of our students.

**Additional context**

A very acceptable version of this would be a pared-down version of a corporate style guide with links away to the full document.

alismith-dev:
## What does this change?

Module: N/A
Week(s): N/A

## Description

Closes https://github.com/CodeYourFuture/syllabus/issues/105.

Adds a code style code.

As discussed on the ticket, the goal was not to be very prescriptive about the style, at least in the earlier parts of the course. Instead we rely on Prettier to do most of the work.

I tried to capture some of the important ideas that are hard enforce through tooling, why style is important and what Prettier does.

[Rendered version](https://github.com/CodeYourFuture/syllabus/blob/code-style-guide/docs/guides/code-style-guide.md).

I also removed the [old style guide](https://syllabus.codeyourfuture.io/guides/coding-standards), since it looked quite out-of-date and contradicted the decisions above.

@ChrisOwen101 
@illicitonion 
@textbook 
@nbogie


docs/guides/code-style-guide.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Using Prettier to format code automatically
+
+Some of the guidelines are so common that there are automatic tools to do it for you! One of these tools is called Prettier. You should have installed this tool when you [set up your VS Code extensions](https://docs.codeyourfuture.io/course-processes/before-the-course/getting-setup#extensions).
+
+Prettier is a tool that rearranges your code (called *formatting*). It follows a set of rules that programmers have agreed make your code easier to read and understand.
+
+Some common problems you may see when running Prettier:
+
+- If you write some code that is invalid, then running Prettier will not work and your code won't be formatted. If this happens, then you will see some red "squiggly" lines under under your code. A common problem is a missing bracket, so try looking for that first
```
ChrisOwen101: Maybe an image would be better here rather than trying to describe it?

alismith-dev: Done in e55c921.

docs/guides/code-style-guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+In this example, there is some extra debugging information that is commented out.
+
+Once you have solved the problem, you should remove large sections of commented out code. When you (or someone else) are reading your code, you want to know only the important pieces. Removing commented out code helps find the relevant code faster and easier.
+
+### Saving old versions with Git
+
+You might be worried about "losing" some experimental code that you want to "save" by commenting it out. However, this can cause confusion about which bits of code still work.
+
+If you want to save some code, remember that you can commit it using Git. It will then be saved it forever. If decide to remove the code later you can delete it and make another commit.
```
illicitonion:
```suggestion
If you want to save some code, remember that you can commit it using Git. It will then be saved it forever. If you decide to remove the code later you can delete it and make another commit. And if you want it back again, you can always look at the deleted code in the git history.
```

docs/guides/code-style-guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Avoid short names
+
+Very short variable names can be difficult to understand since the purpose of the variable can be unclear. They are also difficult remember, especially if there are many similarly named variables. Try to avoid short names or abbreviations.
+
+Here are some examples of **bad** names that you should avoid:
+
+- Single letters like `x` or `y`
+- Abbreviations like `evt` instead of `event`
+
```
illicitonion:
```suggestion
- Abbreviations like `evt` instead of `event`
- Generic names like `array` or `string`

```

docs/guides/code-style-guide.md
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// ✅ This is (probably) better
+function getPercentage() {
+  // ...
+}
+```
+
+In this example, the function calculates a percentage and returns it. Therefore it is good to name it using the "get" verb to show that it returns something.
+
```
illicitonion, line 159:
```suggestion
In this example, the function calculates a percentage and returns it. Therefore it is good to name it using the "get" verb to show that it returns something.

```js
// 🛑 Try to avoid this
function isOldEnough(number) {
  // ...
}

// ✅ This is (probably) better
function isOldEnough(yearOfBirth) {
  // ...
}
```

In this example, we expect the caller of the function to pass a year of birth. The name "number" isn't very clear about what number should be passed. The name "yearOfBirth" makes it more clear, so is a better name. Some functions _are_ very generic, e.g. a function which reverses an array without caring what's stored in it may call its parameter `array`, but try to think of the most specific name that makes sense.

```

**Review by illicitonion — APPROVED** [review]
Looks good to me, thanks!

One thing I thought may be worth adding is something like:

Define your variables with the narrowest scope they can have.

For example in this code:
```js
function findLongestFirstName(fullNames) {
  let longest;
  let firstName;
  for (let i = 0; i < fullNames.length; i++) {
    firstName = fullNames[i].split(" ")[0];
    if (!longest || firstName.length > longest.length) {
      longest = firstName;
    }
  }
  return longest;
}
```

I would recommend declaring `firstName` inside the loop, rather than outside it, so you can better reason about where it's used, but maybe this relies on JS knowledge we don't teach early on...

ChrisOwen101 (approved):

docs/guides/code-style-guide.md
```diff
@@ -0,0 +1,161 @@
...
+Some of the guidelines are so common that there are automatic tools to do it for you! One of these tools is called Prettier. You should have installed this tool when you [set up your VS Code extensions](https://docs.codeyourfuture.io/course-processes/before-the-course/getting-setup#extensions).
+
+Prettier is a tool that rearranges your code (called *formatting*). It follows a set of rules that programmers have agreed make your code easier to read and understand.
+
+You might have a common problem when running Prettier. If you see something like this:
+
+![Invalid code with red highlight](./assets/prettier-error.png)
+
+Then you have written some code that is invalid and Prettier can't format it. A common is a missing bracket, so try looking for that first.
```
timhamrouge: A common what?

docs/guides/code-style-guide.md
timhamrouge: also small typo: 'It will then be saved it forever' in this block

docs/guides/code-style-guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Use `camelCase`
+
+When writing JavaScript, programmers tend to use a convention called *`camelCase`*. This means starting your variable name with a lower case letter, and then every "word" after that starts with an upper case letter.
+
+- `startingLocation`
+- `timeToDestination`
+- `trafficOnRoute`
+
+This convention makes it easier to read variables that have multiple "words".
```
timhamrouge: 'easier to read variable _names_ that have multiple "words"' maybe?

timhamrouge (commented):
just a few little bits

ChrisOwen101:
I've resolved a few of the comments here and separated them into "Essential" and "Advanced" Categories.

docs/guides/code-style-guide.md
```diff
@@ -0,0 +1,223 @@
...
+
+Don't forget that the person reading your code probably has some understanding of the _context_. If your function is named `getCustomer`, it's probably fine to name variables `name`, `age`, etc instead of `customerName`, `customerAge`. You don't need to explain _everything_.
+
+**Naming variables is more of an art than a science**. Often you need some experience reading other people's code before you get really good at it.
+
+:::tip
+When reading other people's code, think about the variable names. If the names are not clear, make sure you understand the code first, then think about what names you would use instead.
+:::
+
+### Good scoping of your variables
```
ChrisOwen101, line 190: @illicitonion @40thieves 

Added your idea here, I think it has value but is a bit more subtle to understand.

I found it hard to describe in simple language _why_ this is important beyond the example given. Happy to take suggestions.

ChrisOwen101: Also - I learnt about line highlighting here! - a nice little markdown touch

40thieves:
Discussed with @ChrisOwen101: I haven't had time to review the comments/changes (thanks everyone!), but happy to merge now and I can submit changes later if needed.

40thieves:
Thanks everyone for taking this over the finish line, and apologies I've only just managed to look through it. And big thanks for catching all my typos 🤦 

Overall I'm happy with where we landed. In my draft I was avoiding some concepts as I felt that they would be overwhelming to JS1 students. But I like the Essential/Advanced structure as a workaround there.

I do wonder if the point about making note of naming while reading other people's code should be bumped into the Essential section? IME there aren't any good rules on naming, so getting trainees thinking about how it feels to read bad names early might give them a leg up in writing better names. I'd be happy to leave the context section in Advanced though.

docs/guides/code-style-guide.md
alismith-dev: Perhaps it is worth adding something along the lines of (adding on to the last paragraph):

It also means that variables can't be used in confusing places. For example, using `firstName` outside of the `for` loop wouldn't make sense. By keeping the variable inside the loop, it can't be used outside.

I feel like there's some better phrasing to be had, but that's the core idea.

PR #160, opened 2021-01-13
Update preparation.md

ultraasi-atanas:
## What does this change?

Module: JS1
Week(s): Preparation - https://syllabus.codeyourfuture.io/js-core-1/preparation/

## Description

Change of the URL for WSL. 

Some students have tried to do the simple install which unfortunately takes them to "install a preview build of Windows 10 (OS build 202062 or higher)". Changing the URL to the manual install, might help them steer in the right direction. 

#southafrica-class1 Sharad Parbhoo Yesterday at 07:41
https://codeyourfuture.slack.com/archives/G01DZBJNR5L/p1610518094054500?thread_ts=1610437308.042000&cid=G01DZBJNR5L

Atanas

### DIFF
--- a/docs/js-core-1/preparation.md
+++ b/docs/js-core-1/preparation.md
@@ -18,7 +18,7 @@ If any of the above or below instructions are unclear, please post in your city 
 
 To install WSL as a Windows 10 user, click on this link and follow the instructions (NOTE: At the section "Install your Linux Distribution of Choice", select Ubuntu 20.04 LTS):
 
-https://docs.microsoft.com/en-us/windows/wsl/install-win10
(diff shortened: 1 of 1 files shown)


ChrisOwen101 (changes requested):
It might be better to recommend them use the primary (and easier) way for now but clearly signpost that if they get the error then they should refer to the secondary instructions.

What do you think?

ultraasi-atanas:
Thanks Chris, I am just hoping to get access to a Windows machine next week to test but at the moment, it does sounds like getting a preview build of Windows 10 sounds like a bit of a big change, but if it has been tested and that's the preferred way, than I am happy

PR #163, opened 2021-01-14
Add Marking Coursework Lesson Session and Coursework

ChrisOwen101:
## What does this change?

Module: JS1
Week(s): 1,2,3

## Description

This PR adds 

- A short 10-15 minute introduction to our style guide with a practical example of how to use it in practice. 
- Assignable coursework to all three weeks of JavaScript Core 1 telling students to give feedback on coursework

The aim of this is to normalise the idea of reading other peoples code well before we reach the point of Final Projects when students may struggle with it.

### DIFF
--- a/docs/js-core-1/week-1/lesson.md
+++ b/docs/js-core-1/week-1/lesson.md
@@ -218,10 +218,11 @@ const preciseAge = 30.612437;
 const roughAge = Math.round(preciseAge); // 31
 ```
 
-### Exercise (15 mins)
+### Exercise (5 mins)
 
-1. Using the variables provided in the exercise calculate the percentage of mentors and students in the group (percentages must be a rounded to the nearest integer)
-2. Using online documentation, what other things can you do with the `Math` library? Pick one thing on your table that you can do other than `Math.round` and prepare an explanation for the rest of the class
+Using the variables provided in the exercise, calculate the percentage of mentors and students in the group (percentages must be a rounded to the nearest integer)
+
+You should then log this number to the console.
 
 #### Expected result
 
@@ -229,6 +230,14 @@ const roughAge = Math.round(preciseAge); // 31
 Percentage students: 65%
 Percentage mentors: 35%
 ```
+
+### Extra Exercise (10 minutes)
+
+_Intructor note: Only complete this exercise if you have enough time in the lesson_
+
+Using online documentation, what other things can you do with the `Math` library?
+
+Pick one thing on your table that you can do other than `Math.round` and prepare an explanation for the rest of the class.
 
 ## Functions
 
--- a/docs/js-core-1/week-1/lesson.md
+++ b/docs/js-core-1/week-1/lesson.md
@@ -327,13 +327,13 @@ This style guide is how we expect you to write code whilst you're at CodeYourFut
 
 [https://syllabus.codeyourfuture.io/guides/code-style-guide](https://syllabus.codeyourfuture.io/guides/code-style-guide)
 
-:::Note
+:::tip
 A style guide doesn't change anything about how your code operates - code that breaks these rules will still work. However it's important so that **other people** can more easily read your code.
 :::
 
 ### Exercise (5 minutes)
 
-_This is a Live exercise. The Teacher should take suggestions from the class on how this code can be improved and add comments themselves_
+_This is a Live exercise. The Teacher should take suggestions from the class on how this code can be improved and add comments themselves. Please read the [teaching notes](./instructors#giving-feedback-to-each-other)._
 
 Next, let's see an example of some code that we could suggest some improvements to.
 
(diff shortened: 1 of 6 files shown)


szemate:
The addition looks very useful, but with this change, I don't think it's possible to fit all the exercises from the syllabus into one day with proper explanation, it's too much. I'd mark the second exercise of "Numbers as decimals" about using online documentation as optional/extra, I think introducing the style guide is more important at this point, and it would save 10 minutes.

docs/js-core-1/week-1/mentors.md
```diff
@@ -16,6 +16,16 @@ You can find the preparation material [here](../preparation).
...
 
+## Delivery Notes
+
+### Giving Feedback to Each Other
+
+For this exercise, you should open the Pull Request and share you screen. Illicit answers from the students based on what they've just read.
+
+You should only type what they say, try to let them take the lead.
+
+**Important**: Please delete the comments after you have finished the exercise so that another teacher can use it!
```
ChrisOwen101:
```suggestion
**Important**: Please delete the comments after you have finished the exercise so that another teachers can use it!
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -316,6 +326,33 @@ function createGreeting(name, age) {
 1. Write a function that returns the year someone is born given their age as input
 2. Using the answer from step 1, write a function that takes someone's name and age as input and returns a string that states the person's name and year they were born in a sentence
 
+## Giving Feedback to Each Other
+
+A critical part of being developer is reading other people's code and giving good feedback and improvements. This can happen even when you don't know a lot of code - everybody has unique ideas and insights into how to make code better!
+
+Let's spend the next five minutes reviewing CodeYourFuture's Style Guide.
+
+This style guide is how we expect you to write code whilst you're at CodeYourFuture - each company has one and it's really important you stick to it to make sure we all write code that other people can understand.
```
illicitonion:
```suggestion
This style guide is how we expect you to write code whilst you're at CodeYourFuture - most companies have one and it's really important you stick to it to make sure we all write code that other people can understand.
```

docs/js-core-1/week-1/mentors.md
```diff
@@ -16,6 +16,16 @@ You can find the preparation material [here](../preparation).
 
 Be sure to assign all of the tasks to your students several days before you aim to teach.
 
+## Delivery Notes
+
+### Giving Feedback to Each Other
+
+For this exercise, you should open the Pull Request and share you screen. Illicit answers from the students based on what they've just read.
```
illicitonion:
```suggestion
For this exercise, you should open the Pull Request and share you screen. Elicit answers from the students based on what they've just read.
```

illicitonion (approved):
I'm a bit concerned about unhelpful/misleading/incorrect comments...

Will anyone be checking over the comments left between trainees to correct potentially misleading ones? If so, how should they be handled?

Will it be clear to trainees which comments are from mentors and which are from other trainees?

Do we perhaps want to add some guidance in the "pick two PRs" instruction to say something like "pick two PRs **from people in your class**"? It's probably more likely that a trainee will recognise the role of someone in their class, than determine whether some trainee from another city is a volunteer or trainee...

docs/js-core-1/week-1/mentors.md
ChrisOwen101: Ha! I blame your Github handle for this one!

ChrisOwen101:
@illicitonion 

I completely understand your concerns but I think it might be necessary risk. 

I'd consider it in the same kind of way that we encourage students to work together outside of volunteer time - they might have conversations that are unproductive or misleading but it's the risk we take considering how beneficial peer learning can be. 

Of course, different challenges here - miscommunication is easier,  it can often be unidirectional, missing the tone etc.

>Will it be clear to trainees which comments are from mentors and which are from other trainees?

Most of the feedback they get will be from their Education Buddy so they'll know them well.

illicitonion:
> I'd consider it in the same kind of way that we encourage students to work together outside of volunteer time - they might have conversations that are unproductive or misleading but it's the risk we take considering how beneficial peer learning can be.

I think the difference here, though, is how authoritative the advice seems. I feel (hope?) more weight is given by trainees to technical advice given to them by volunteers than peers. In peer-driven sessions, I think it's obvious to the trainees where the advice is coming from. In code review, this is less obvious.

> > Will it be clear to trainees which comments are from mentors and which are from other trainees?
> 
> Most of the feedback they get will be from their Education Buddy so they'll know them well.

Are we sure that's the case? In the HTML/CSS class, I've seen folks asking for all-hands-on-deck to help mark homework...

To be clear, I'm not suggesting we _don't_ make this change, just wondering if there are ways we can make the volunteer vs peer distinction clearer in individual reviews. "Reviewer isn't your edu buddy means it's peer feedback" is a reasonable solution to this concern, assuming it's consistently applied.

PR #164, opened 2021-01-15
Add "Using let, const and var" to the Style Guide

ChrisOwen101:
## What does this change?

The Code Style Guide

## Description

We should probably have a policy of let, const and var for our students. 

This might be a little opinionated but I think it's the right way to go.

[Rendered version](https://github.com/CodeYourFuture/syllabus/blob/docs/guides/code-style-guide#using-let-const-and-var)


illicitonion (approved):
Works for me!

PR #166, opened 2021-01-21
Terminal Lesson

ChrisOwen101:
## What does this change?

Module: Git/Terminal
Week(s): 2

## Description

**Note**: This is Self-taught lesson that will be included as part of the coursework for a PD week.

This lesson covers the Learning Objectives in the Lesson Plan. The LO's are achieved in

- Define the differences between a Terminal and a GUI and describe the benefits of both **(In Lesson)**
- Open a terminal window **(In Lesson)**
- Using the terminal navigate to a specified folder (using `cd`) **(In OpenClassroom Course)**
- Move up a level in a folder directory (using `cd ..`) **(In OpenClassroom Course)**
- Cancel a running task in the terminal (using `ctrl+c`) **(In OpenClassroom Course)**
- View all the files of the current directory they are in (using `ls`) **(In OpenClassroom Course)** 
- Use arguments to augment the commands learnt in the lesson (e.g. `ls -all`) **(In OpenClassroom Course)**
- Clear the terminal window (using `clear`) **(In CodingTrain Course)**
- Use autocomplete when in the terminal to complete commands (using the Tab key) **(In CodingTrain Course)**
- Find which directory you're currently in (using `pwd`) **(In OpenClassroom Course)**

All of this is tied together in an assessment piece which is peer-reviewed. 

https://github.com/CodeYourFuture/Terminal-Coursework-Week1

This lesson stays purposefully quiet limited and is our first foray into entirely self-taught content. Students are signposted at several points to ask for help when they need it. 

@illicitonion @timhamrouge


docs/git/terminal/lesson.md
```diff
@@ -0,0 +1,216 @@
...
+
+Many of the tasks that you complete using a GUI can also be done using the command line. For example,
+
+- Creating a file
+- Creating a folder
+- Looking at the contents of a file
+- Connecting to the internet
+- Opening programs
+
+You can write `commands` to your computer by wring text after the `$` sign in your terminal window and pressing `enter`.
```
illicitonion: More recent versions of macOS default to `zsh` which uses a `%` symbol instead of a `$` symbol... I'm not sure how to best convey that without being confusing... Maybe just something like:
```suggestion
You can write `commands` to your computer by wring text after the `$` sign in your terminal window and pressing `enter`. (Some terminals use the `%` or `#` symbol instead of `$` - they work exactly the same).
```

docs/git/terminal/lesson.md
```diff
@@ -0,0 +1,216 @@
...
+
+<Tabs
+defaultValue="exercise"
+values={[{ label: "Exercise", value: "exercise", },{ label:"Answer", value: "answer", },]}>
+<TabItem value="exercise">
+
+:::tip Exercise
+Make the command line do something!
+
+It's really really hard to break something in the terminal by mistake.
```
illicitonion: I'm slightly wary of saying this... Something like `sudo rm -rf /` is really easy to do, and if they're maybe going to Google random stuff to run in their terminal, it's not impossible they'll run into something like this...

illicitonion (commented):
In general I think this looks great, and OpenClassroom seems like a great thing to lean on :)

A few comments on the coursework:
* The README uses the term "find" a lot, like "find the `script.js` file" - I'm not sure "finding" is particularly well defined... Maybe we could be more explicit, by saying something like "change directory to the directory containing `script.js`?
* For the last question, the quiz one, I don't think it's obvious how the trainees should open the quiz - as part of a terminal class, I'm kind of expecting to need to use the terminal to open it, but I don't think we've explained that `code MyDocuments/Quiz` or `code QUIZ.md` will open VS Code? Maybe we should either explicitly tell them to open the coursework repo in VS Code and complete the quiz (so they're not thinking about how to do so in the terminal), or teach them how to open VS Code from the terminal?

And on the individual quiz questions...
* I'm not sure what the answer to the first question is... I'm guessing the last one, but I'm also not convinced it's incorrect?

I also made a couple of small fix-up PRs to the coursework repo: https://github.com/CodeYourFuture/Terminal-Coursework-Week1/pull/1 and https://github.com/CodeYourFuture/Terminal-Coursework-Week1/pull/2 and one to the solutions repo: https://github.com/CodeYourFuture/Terminal-Coursework-Week1-Solution/pull/1

docs/git/terminal/lesson.md
ChrisOwen101: Fair enough! I'll replace it with a list of words

40thieves:
Late to the party, but this looks good 👍  I like the OpenClassrooms lesson, it's well written for this sort of thing.

Thinking about the LOs, one of the "dependencies" I think we're missing is usage of paths in `import`/`require` statements. This lesson does touch on that a bit, but it doesn't really discuss what `./` means, and there's no particular connection to JS.

I see 2 options:

1. Add some additional coursework to the terminal lesson about interacting with paths in JS. This could be as simple as running some things in the REPL?
2. It doesn't quite fit into any of the other lessons, so we could just add content [in the React lesson](https://syllabus.codeyourfuture.io/react/week-1/lesson#importingexporting-components) when we use paths for the first time. However, week 1 of React is pretty jam-packed as it is.

Or I could be overthinking it, and we could assume that students will "get it" as soon as they see paths.

illicitonion:
> Or I could be overthinking it, and we could assume that students will "get it" as soon as they see paths.

Specifically for `./`, the majority of trainees I've worked with haven't understood what it meant without having it explicitly explained, and I don't think the content in here will significantly change that.

PR #168, opened 2021-01-22
Functions expressions exercises

Dedekind561:
## What does this change ?

Module: js-core-1 & js-core-2
Week(s): 1 & 2

## Description

This PR moves expressions from week-2 into week-1 and then updates the exercises for:
-  statements and expressions, 
- functions
- the first set of exercises for **booleans** in week-2.

The PR also removes functions calling other functions from the end of week-1.

Please check out the links below to see the rendered versions of the updated lessons:

  [Rendered version of js-core-1/week-1/lesson.md](https://github.com/CodeYourFuture/syllabus/blob/functions-expressions-exercises/docs/js-core-1/week-1/lesson.md)
  [Rendered version of js-core-1/week-2/lesson.md](https://github.com/CodeYourFuture/syllabus/blob/functions-expressions-exercises/docs/js-core-1/week-2/lesson.md)
  
@syllabus-team


docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
+
+f) What will the variable `messageForAli` evaluate to when the code above is executed ?<br />
+g) What will the variable `curiousMessage` evaluate to when the code above is executed ?
+
+Hint: Think carefully about the parameters and arguments for `getMessage` for part g)
+
+
+### Exercise (10 minutes)
+
+1. Swap your laptop with your neighbouring pair and review each other's code - what can be improved? Is the code readable?
```
illicitonion: Do we want to be more explicit about what code we're talking about here? Maybe something like:
```suggestion
1. Swap your laptop with your neighbouring pair and review each other's code for the `formatPenceToPounds` and `increaseByHalf` exercise - what can be improved? Is the code readable?
```

Also, do we want to suggest a concrete way of doing this over Zoom? Something like "Pair up, and send your code to your pair over Slack"?

docs/js-core-1/week-1/lesson.md
```diff
@@ -25,6 +23,8 @@ By the end of this class, you should be able to:
 - List and describe the different primitive data types in JS
```
illicitonion: (I realise you didn't change this, but...) We don't actually teach this at all, we just define it in a glossary at the bottom... Maybe worth removing, and just not introducing the term "primitive"? We're already introducing a _lot_ of definitions i this lesson...

docs/js-core-1/week-2/lesson.md
```diff
@@ -158,7 +85,7 @@ The `>` symbol in the expression is a **comparison operator**. Comparison operat
 !== not the same value and type
 ```
 
-You might see people use two other comparison operators. They are valid, but are the evil twins of `===` and `!==` and it is generally considered bad practice because of the bugs the can cause.
+You might see people use two other comparison operators. They are valid, but are the evil twins of `===` and `!==` and it's generally considered bad practice because of the bugs they can cause.
```
illicitonion:
```suggestion
You might see people use two other comparison operators, `==` and `!=`. They are valid, but are the evil twins of `===` and `!==` and it's generally considered bad practice because of the bugs they can cause.
```

docs/js-core-1/week-2/lesson.md
```diff
@@ -176,43 +103,86 @@ You might see people use two other comparison operators. They are valid, but are
...
 
-const thisNumber = // TODO
-const thatNumber = // TODO
-assert(thisNumber === thatNumber);
+// You will need to research falsy values in order to work out the next 2 problems below!
 
-const thisArray = [1, 2, 3, 4, 5];
-const thatArray = [1, 2, 3, 4, 5];
-assert(thisArray === thatArray);
+const mysteryString = FILL_ME_IN;
```
illicitonion: I think we agreed in the last meeting to cut out the flasey stuff, and to remove the array equality exercise and have the teacher explain the array equality stuff instead?

illicitonion (commented):
Looking great! Thanks for putting this all together!

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+A statement is some code that typically carries out an instruction. You have encountered statements already:
+
+```js
+const greeting = "Hello Code Your Future!";
+```
+
+The above piece of code is a **variable declaration**, which is also a statement. 
+This statement is just an instruction to store the value `"Hello Code Your Future!"` in the variable `greeting`.
+
+Statements end in a semi-colon, though they can be omitted in JavaScript.
```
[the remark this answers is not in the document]
alismith-dev: We did discuss that understanding of what "evaluating an expression" means is a useful tool for instructors, and I do think we refer to things like "if statements" fairly often. So having a definition is perhaps a good idea.

I would lean towards moving the statements section below the expressions section. I think it may be easier to define statements in context of expressions: "If expressions are single units of JavaScript that the engine can evaluate, statements can contain one or more different expressions, and are executed by the engine to perform an operation" ([source](https://flaviocopes.com/javascript-statements/))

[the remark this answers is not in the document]
alismith-dev: Also, what do you think about leaving out the point about semi-colons being optional?

Benefits:

- Prettier enforces semis by default, so trainees shouldn't have to worry about them
- JS' automatic semi-colon insertion is infamously weird, and discussing it will be distracting

Drawbacks:

- Some volunteers (like myself) are in the habit of leaving off semis, so it tends to come up when live-coding

I am a bit on the fence, but lean towards leaving it out.

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+
+There are other different types of statements that we will learn about in the coming weeks.
+
+### Expression
+
+Any unit of code that can be evaluated to a value is known as an **expression**.<br />
+We say that an expression _evaluates to_ a value.
```
alismith-dev:
```suggestion
Any unit of code that can be evaluated to a value is known as an **expression**.

We say that an expression _evaluates to_ a value.
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+
+In your terminal, run the command `node` and then enter a line of code from the code below and then run enter. Do this for each line of code in the list.
+Think about the following questions each time:
+
+What is the output in the terminal each time ?
+Is there anything you didn't expect ?
+
+Can you work out which of the lines of code below are **expressions** and which are not ?
+
+_Remember it can only be an expression if you can place it on the right hand side of an assignment._
```
alismith-dev, line 288: This feels a bit out of place, since it's the first time it's come up.

What do you think about adding a (small) section comparing expressions and statements, where we could show a code snippet with the differences?

docs/js-core-1/week-1/lesson.md
```diff
@@ -239,6 +233,106 @@ Using online documentation, what other things can you do with the `Math` library
...
+if (true) {}
+
+const hoursInADay = 24
+
+hoursInADay * 7
+
+`My favourite colour is ${favouriteColour}`
+```
+
+(To exit the node REPL, you have to click Ctrl+d or Cmd+D on Mac)
```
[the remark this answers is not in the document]
alismith-dev:
```suggestion
(To exit the node REPL, you have to press Ctrl+d)
```

Apologies, I feel like I'm overly nitpicking here.

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
+To use a function then we we need to **call** it ( some people will say **invoke** )
 
 ```js
 const result = double(2);
 
 console.log(result); // 4
 ```
 
-The input given to a function is called a **parameter**.
+The function `double` is being invoked by placing parathenses after the name of the function like this `double()`
```
alismith-dev: Given that we say "call" as the main verb above, and most volunteers in my experience use that, could we change to:

```suggestion
The function `double` is being called by placing parathenses after the name of the function like this `double()`
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
+
+🔑
+The input given to a function is called a **argument**
+
+🔑 A **parameter** is a variable that is used to refer to an **argument**
```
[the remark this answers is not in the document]
alismith-dev: What do you think about rephrasing Daniel's suggestion to:

> In the example above, we're saying that inside the function `double`, there is a variable called `number`, but we don't know what the value of the variable is until someone calls the function, at which point they will give a value for that variable.

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
 function add(a, b) {
   return a + b;
 }
 ```
 
-When you write a function (sometimes called _declaring a function_) you assign names to the parameters inside of the parentheses (`()`). Parameters can be called anything.
+In the function `add`, the parameters are `a` and `b`.
 
-This function is exactly the same as the on above:
+If we call the function like this `add(10,32)` then we're passing 2 arguments (or inputs) into the function : `10` and `32`.
```
alismith-dev:
```suggestion
If we call the function like this `add(10, 32)` then we're passing 2 arguments (or inputs) into the function : `10` and `32`.
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
 
-2. Add a comment above your function to explain what it does
-3. Call your function and run your script
-4. What's the difference between a `return` and `console.log`?
-5. When would you choose to use functions over the way we have been scripting so far?
+1. Create a function called `formatPenceToPounds` that should
 
-### Exercise (10 minutes)
+- take a number as an input ( representing an amount of money in pence )
+- return a string starting with a`£` sign and the input number converted to pounds.
```
alismith-dev:
```suggestion
- return a string starting with a `£` sign and the input number converted to pounds.
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
 function add(num1, num2) {
   return num1 + num2;
 }
 ```
 
-### Exercise (20 minutes)
+### Exercises (20 minutes)
 
-1. In pairs, design and create a function that:
+Try creating a new file when solving the each of the questions below:
```
alismith-dev:
```suggestion
In a new file (`js1-week1-exercises.js`), try solving each of the questions below:
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
+
+- what the parameter name should be
+- creating good names for any variables you use
+- don't use `console.log` inside the body of the function
+
+Call `formatPenceToPounds` a few times with different inputs and `console.log` the outputs in order to check your function. 🙂
+
+
+---
+<br />
```
alismith-dev:
```suggestion
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
 
-Functions are very powerful.
+Call `increaseByHalf` a few times with different inputs and `console.log` the outputs in order to check your function is working properly.
 
-- You can write more than one line of code inside of functions.
-- You can use variables inside of functions.
-- You can call other functions inside of functions!
+
+Now declare a variable called `pocketMoney` and assign to it a value of `413`.<br />
+Someone's been generous and decided to increase your `pocketMoney` by 50% 😎
```
alismith-dev:
```suggestion
Now declare a variable called `pocketMoney` and assign to it a value of `413`.

Someone's been generous and decided to increase your `pocketMoney` by 50% 😎
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
+
+Using the functions `increaseByHalf` and `formatPenceToPounds` only, find the new amount of pocket money as a string in pounds with a `£` symbol at the front.
```
alismith-dev, line 421:
```suggestion
Using the functions `increaseByHalf` and `formatPenceToPounds` only, find the new amount of pocket money as a string in pounds with a `£` symbol at the front.

Then use `console.log` the output in order to check your function.
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
+
+
+---
+<br />
```
alismith-dev:
```suggestion
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
 
-### Exercise (20 mins)
+The functions above behave similarly but _differ_ in some important ways.
+
+In groups, study the code above and then answer the following questions:
+
+a) How many times is the function `printMessage` called ?<br />
+b) How many times is the function `getMessage` called ?<br />
+c) What is the parameter name for the function `printMessage` ?<br />
+d) What is the parameter name for the function `getMessage` ?<br />
```
alismith-dev:
```suggestion
a) How many times is the function `printMessage` called ?
b) How many times is the function `getMessage` called ?
c) What is the parameter name for the function `printMessage` ?
d) What is the parameter name for the function `getMessage` ?
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
+
+Someone then creates 2 further variables `messageForAli` and `curiousMessage`:
+
+```js
+const messageForAli = getMessage('Ali');
+
+const curiousMessage = getMessage(42);
+```
+
+f) What will the variable `messageForAli` evaluate to when the code above is executed ?<br />
```
alismith-dev:
```suggestion
f) What will the variable `messageForAli` evaluate to when the code above is executed ?
```

docs/js-core-1/week-1/lesson.md
```diff
@@ -251,80 +345,143 @@ function double(number) {
...
+g) What will the variable `curiousMessage` evaluate to when the code above is executed ?
```
alismith-dev, line 467: I wonder if it's worth adding a point about how to find out what the answers are?

I understand that one of the goals is to reduce confusion about `console.log`, but I think this is a good opportunity to show that it should be used for debugging purposes.

docs/js-core-1/week-2/lesson.md
```diff
@@ -108,46 +41,40 @@ let codeYourFutureIsGreat = true;
 let thisIsATerribleClass = false;
 ```
 
-### Exercise (10 mins)
+We can use boolean values to make decisions in our code based on certain conditions, as we shall later on.
```
alismith-dev:
```suggestion
We can use boolean values to make decisions in our code based on certain conditions, as we shall see later on.
```

docs/js-core-1/week-2/lesson.md
```diff
@@ -108,46 +41,40 @@ let codeYourFutureIsGreat = true;
...
-  return "No bool, not cool.";
-}
+For example, 
 
-boolChecker(true);
-```
-
-3. As a class, can you step through the function and explain what each line does?
+`42 === 42` will evaluate to `true`.
+You can check this by logging the expression `42 === 42` in the console.
```
alismith-dev: I wonder if we should change "console" to "Node REPL"?

I think this could be interpreted as adding `console.log(42 === 42)` to a file and running it. Which isn't wrong per-se, but I think the intention is to run a quick check in the repl. Adding more logging to their working file may also get confusing.

docs/js-core-1/week-2/lesson.md
```diff
@@ -108,46 +41,40 @@ let codeYourFutureIsGreat = true;
...
+You can also check this out by logging the expression `42 === 50` in the console.
 
-In the previous exercise, you used an **expression** that returns a `boolean` value. This was possible because of the **comparison operator** `===`. Using a comparison operator will always return a `boolean` value.
+Using a comparison operator will always return a `boolean` value.
 
-### Assert
+### `checkIsTrue`
 
-We will also be using a new keyword to make sure the comparisons we do below return `true`. This is the `assert`.
+We have created a function `checkIsTrue` that checks if an **expression** evaluates to `true`or `false`. For the time being, you don't need to know how the function is implemented - you will only need to understand what the function does.
```
alismith-dev: Would this read cleaner if we started with something like this?

```suggestion
Let's pretend that we have created a function `checkIsTrue` that checks if an **expression** evaluates to `true` or `false`. For the time being, you don't need to know how the function is implemented - you will only need to understand what the function does.
```

docs/js-core-1/week-2/lesson.md
```diff
@@ -176,43 +103,86 @@ You might see people use two other comparison operators. They are valid, but are
...
+
+const nums = [1, 2, 3, 4, 5];
+const numsCopy = FILL_ME_IN;
+checkIsTrue(nums === numsCopy);
+
+// Don't change anything below this point
+var FILL_ME_IN;
+function checkIsTrue(value1) {
+  if (value1) {
+    console.log("\x1b[32m%s\x1b[0m","Yes, this expression evaluates to true! ✅");
```
alismith-dev: Should we add a comment to (briefly) explain what the weird characters are about? Or perhaps we should consider removing them entirely?

docs/js-core-1/week-2/lesson.md
```diff
@@ -176,43 +103,86 @@ You might see people use two other comparison operators. They are valid, but are
...
+checkIsTrue(organisation.startsWith('Code'));
+checkIsTrue(organisation.includes(' your '));
+checkIsTrue(organisation.endsWith('Future!'));
 
-const name1 = // TODO
-const name2 = // TODO
-assert(name1 === name2);
+const firstName = FILL_ME_IN;
+const lastName = FILL_ME_IN;
+checkIsTrue(firstName[0] === lastName[0]);
```
alismith-dev: I think this is the first time trainees would encounter string indexing? Should we consider removing these lines and covering that explicitly? It can be confusing as trainees tend to think only of arrays when seeing indexing syntax.

docs/js-core-1/week-2/lesson.md
```diff
@@ -176,43 +103,86 @@ You might see people use two other comparison operators. They are valid, but are
...
+  } else {
+    console.log("\x1b[31m%s\x1b[0m","Sorry, this expression evaluates to false! ❌");
+  }
+}
 ```
 
 The array comparison doesn't work because JavaScript comparison only works as expected on `number`s, `string`s, and `boolean`s.
 
-You need to be aware of this, and its possible for you to go deeper into array comparisons outside of the lesson, though generally remember that only `number`s, `string`s, and `boolean`s work with equality.
+You need to be aware of this, and you can go deeper into outside of the lesson, but generally remember that only `number`s, `string`s, and `boolean`s work with equality.
```
alismith-dev:
```suggestion
You need to be aware of this, and you can go deeper into outside of the lesson, but generally remember that only `number`s, `string`s, and `boolean`s work with equality comparisons.
```

Also (and this is probably a separate conversation) I wonder if we should explicitly address comparing arrays/objects in JS2?

docs/js-core-1/week-2/lesson.md
```diff
@@ -176,43 +103,86 @@ You might see people use two other comparison operators. They are valid, but are
...
+
+---
+
+3. Create a function called `isEven` that should
+- take a number as an input
+- return a **boolean** indicating if the number is even or odd
+
+Remember to avoid using `console.log` _inside_ the function itself
+
+E.g. `isEven(100)` should return `true`<br />
```
alismith-dev:
```suggestion
E.g. `isEven(100)` should return `true`
```

alismith-dev (commented):
Looking good! Thanks for all the hard work here 🙂 

I've left a few comments on formatting, hopefully you don't mind me nitpicking. I think they may be artefacts from copy/pasting from Notion?

40thieves:
Sorry, remembered another thing: it looks like there's a 16,000+ line diff on `package-lock.json` - was that intentionally included?

docs/js-core-1/week-2/lesson.md
Dedekind561: I'll make a list of these sorts of things that should be raised as other issues @40thieves

docs/js-core-1/week-2/lesson.md
Dedekind561: I personally don't think it is that tricky but am also happy to move it to an area where indexing is covered more explicitly

docs/js-core-1/week-1/lesson.md
```diff
@@ -92,13 +92,13 @@ We can use a _variable_ to create a reference to a value. Variables can be thoug
 Before you use a variable in a JavaScript program, you must declare it. Variables are declared with _let_ and _const_ keywords as follows.
 
 ```js
-let greeting = "Hello world";
+let greeting = 'Hello world';
```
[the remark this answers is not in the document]
Dedekind561: Yeah I've changed those now using the default settings in Prettier. Do you think it is worth adding a `.prettierrc` file

illicitonion (approved):
This looks great, thanks so much for working on it!

PR #177, opened 2021-02-05
Git and the Command Line

ChrisOwen101:
## What does this change?

Module: JS2-2 and Git-3

## Description

This PR adds a lesson on Git and the Command Line.

@illicitonion


docs/git/overview.md
```diff
@@ -14,3 +14,4 @@ Each week, you're going to use Git to get a copy of your Coursework exercises, t
 | --------------------------- | ----------------------- | ------------------------------------------ |
 | [Week 1](./index.md)        | Github & Github Desktop | [Git Coursework](./homework)               |
 | [Week 2](./terminal/lesson) | The Terminal            | [Terminal Coursework](./terminal/homework) |
+| [Week 2](./cli/lesson)      | Git CLI                 | [Terminal Coursework](./cli/homework)      |
```
illicitonion:
```suggestion
| [Week 3](./cli/lesson)      | Git CLI                 | [Terminal Coursework](./cli/homework)      |
```

docs/git/cli/lesson.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Coursework
+
+You can find this weeks coursework [here](./homework).
+
+## Feedback
+
+It's really helpful if you give us feedback for this lesson. This helps us improve it for future students!
+
+<Feedback module="Git" week="Week 2" />
```
illicitonion:
```suggestion
<Feedback module="Git" week="Week 3" />
```

docs/git/cli/lesson.md
```diff
@@ -0,0 +1,81 @@
...
+import Feedback from "@theme/Feedback";
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+## Learning Objectives
+
+By the end of this class, you should be able to:
+
+- To _clone_ a repository using Git CLI
+- To _commit_, _fork_ a repository, and _push_ changes using Git CLI
```
illicitonion: They won't really be forking from the CLI, they'll still be forking through the web UI, but then more be setting up their local clone to use their fork... I'm not sure of a concise way to express that, though...

docs/git/cli/lesson.md
```diff
@@ -0,0 +1,81 @@
...
+- To view Git's _history log_ using Git CLI
```
illicitonion, line 18: I don't think I see `log` anywhere in the course material?

(I definitely think we should introduce `log`, but this also means we need to explain navigating the `less` pager, so paging up/down, and `q` to exit - maybe even `/`/`?` for searching... Probably not...)

Also, do we want to introduce `git diff`? If we're introducing navigating `less`, we might as well, I think...

It looks like `git-it` introduces `git diff` in the "Commit To It" step, but it doesn't tell you how to quit out of `git diff`, or how to navigate around it, which is pretty concerning...

docs/git/cli/lesson.md
illicitonion: I think we should have an LO about `git status` too... git-it tells them to use `git status`, but doesn't really explain much about it, but that's probably enough for this class.

Maybe an LO along the lines of:
* See what uncommitted changes you have with `git status`

docs/git/cli/homework.md
```diff
@@ -0,0 +1,58 @@
...
+
+https://syllabus.codeyourfuture.io/git/terminal/lesson
+
+## 2) Git and Github for Poets (3 Hours)
+
+This video series gives you a good introduction to lots of different Git concepts on the command line. You might recognise the presenter from the Terminal week!
+
+https://www.youtube.com/watch?v=BCQHnlnPusY&list=PLRqwX-V7Uu6ZF9C0YMKuns9sLDzK6zoiV
+
+Watch all 11 videos in the series, taking notes when you don't understand something.
```
illicitonion: This includes merge conflicts - do we maybe want to stop after video 7 or 8?

illicitonion (commented):
Looks like a good shape :) I'm a little weirded out by Git It, partially because I'm not sure why it's an electron app rather than just a website, partially because it seems a little light on detail, and particularly light on edge cases (e.g. by telling people to run `git diff` without telling them how to get out of `less`; I'm also a little concerned that we don't have a "gotcha: What if you run `git commit` and forget `-m` and end up stuck in vim"...) - but if we can rely on pre-written material, that would be lovely, so I'm certainly happy to give it a go (and contribute improvements upstream)!

docs/git/cli/lesson.md
```diff
@@ -0,0 +1,87 @@
...
+
+### You feel comfortable everywhere
+
+A big advantage for using Git CLI is that you work comfortably everywhere, the experience of working on your local computer, a colleague computer or even a remote machine with a different OS feels more or less the same (minus your personal configurations perhaps).
+
+### Fully utilizing Git
+
+The big downside of using GUI for Git is that it exposes you to a relatively small interface (whatever part the GUI creator deemed important), this basically means you are not fully interacting with Git and are therefore not familiar with (at least) some of its capabilities.
+
+There are some really good developers that used a certain GUI and were not even aware that Git has a staging area. When using the Git CLI you are using the entire Git interface and are therefore able to see the bigger picture, this makes you much more comfortable with Git and will almost certainly expose you to some useful tools you might not have used otherwise.
```
illicitonion: I know you're quoting here, but... Given we haven't taught our trainees about the staging area yet, "[some developers] were not even aware that Git has a staging area" feels like quite a weird thing to say... Maybe we could introduce some concept that they haven't interacted with because it's hard to do from a GUI here? (Though nothing obviously comes to mind...)

docs/git/cli/lesson.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Git CLI gives you the power of the command line
+
+We often try to reuse & build on top of existing tools and this is no different since you’re already using the command line, you are able to compose your Git flow with many useful command-line tools. From looping over each commit and running tests on it, through easily adding files using FZF to using Ag on your logs — the only limit here is your creativity.
```
illicitonion:
```suggestion
We often try to reuse & build on top of existing tools and this is no different since you’re already using the command line, you are able to compose your Git flow with many useful command-line tools. From looping over each commit and running tests on it, through easily adding files using [FZF](https://www.youtube.com/watch?v=QeJkAs_PEQQ) to using [Ag](http://conqueringthecommandline.com/book/ack_ag) on your logs — the only limit here is your creativity.
```

I'm not convinced these are the _best_ resources we could be linking to, but I feel like we probably should be linking to something, rather than just saying the names of these tools...

docs/git/from-desktop-to-cli-cheatsheet.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Committing Changes
+
+Committing changes to your repository can be done by using this command
+
+```sh
+git commit -m "YOUR MESSAGE HERE"
+```
+
+The `-m` means that after you are going to write your commit message.
```
illicitonion:
```suggestion
The `-m` means that after you are going to write your commit message. Make sure to put your commit message in quotes! You can press enter to add new lines in your message, too.

Some characters are treated specially in the terminal, and we can tell the terminal not to treat them specially by _escaping_ them with a backslash. For example, if you want to use a quote character inside your message, you need to escape it, otherwise the terminal will think your quote is the end of the message:

```sh
git commit -m "Use \"s instead of 's" 
```

One of the other characters that needs escaping is the exclamation mark - if you want to use an exclamation mark in a commit message, make sure to escape it:

```sh
git commit -m "Make it prettier\!"
```
```

File: docs/git/from-desktop-to-cli-cheatsheet.md [hunk 9]
```diff
@@ -0,0 +1,85 @@
...
+| Github Desktop                                               | Git CLI                                             |
+| ------------------------------------------------------------ | --------------------------------------------------- |
+| ![Commiting Changing with Github Desktop](assets/commit.png) | ![Commit files with Git CLI](assets/commit-cli.gif) |
+
+### Pushing Changes to Github
+
+Finally, pushing your changes to Github can be completed by using
+
+```
+git pull
```
illicitonion:
```suggestion
git push
```

docs/git/cli/homework.md
```diff
@@ -0,0 +1,58 @@
+---
+id: homework
+title: Homework
+sidebar_label: Homework
+---
+
+## 1) Read the Lesson Page Content (60 minutes)
+
+Before starting this coursework, read through the lesson and complete the exercises.
```
illicitonion:
```suggestion
Before starting this coursework, read through the lesson and read the cheat-sheet.
```

There aren't any exercises in the lesson, but we should make sure they read the cheat-sheet, rather than just acknowledging it's there :)

illicitonion (commented):
Looking good, loving the cheat-sheet!

docs/git/from-desktop-to-cli-cheatsheet.md
illicitonion: Ugh github isn't great at suggestions that include code blocks :D If you try to edit this message, you should see what I was trying to suggest...

PR #207, opened 2021-03-15
Add diagrams showing chains of commits

### LINKED ISSUE #146: Git: Add a "chain of commits" diagram
Module(s): Git
Week(s): Week 1

We don't really talk about history much in the class - we mostly just show it in the GitHub Desktop UI.

It would probably be good to introduce a visualisation showing history as a (linear) chain of commits, a visualisation showing that when we make local commits our chain is a superset of the remote's chain, and when we push we're effectively copying the head of the chain over and attaching it to the remote chain.

illicitonion:
## What does this change?

Module: Git
Week: 1

## Description

Adds diagrams showing history as a chain of commits.

  [Rendered version](https://github.com/CodeYourFuture/syllabus/blob/adding-git-commit-chains/docs/git/index.md)

Fixes #146


ChrisOwen101:
Great work - thanks Daniel!

PR #234, opened 2021-04-12
Simplify React setup instructions by using the CYF template repo

alismith-dev:
## What does this change?

Module: React (well it's a guide, but whatever)
Week(s): 0

## Description

Switches the React set up instructions to use the CYF Create React App template: https://github.com/CodeYourFuture/cra-template. This is essentially identical to the base template, but installs/configures the `stop-runaway-react-effects` dependency. We can then remove it from the instructions too.

I ran the command and it appears to install the latest React/react-scripts so I think it'll be up-to-date.

@DomVinyard


DomVinyard (approved):
Great, this is way more concise 👍

PR #239, opened 2021-04-24
Update Node week2 lesson as body-parser is no longer required

szemate:
## What does this change?

Module: Node
Week(s): 2

## Description

Since Express v4.16.0 the JSON middleware has been part of the core Express API and the `body-parser` Node module is no longer required.


illicitonion (approved):
Thanks!

PR #245, opened 2021-05-09
json() middleware now part of express.

MarkFarmiloe:
## What does this change?

Module: DB
Week(s): 3

## Description

Update lesson to reflect that the json middleware is now part of express.

 [Rendered version](https://github.com/Markfarmiloe/syllabus/blob/master/docs/db/week-3/lesson.md)

Chris Owen

### DIFF
--- a/docs/db/week-3/lesson.md
+++ b/docs/db/week-3/lesson.md
@@ -80,19 +80,12 @@ In the following, we will add a new API endpoint to create a new hotel in the ta
 INSERT INTO hotels (name, rooms, postcode) VALUES ('New Hotel', 5, 'ABC001');
 ```
 
-As we create a new record in the database, we will add a new POST endpoint in the `cyf-hotels-api` project from last class. Moreover, we need to be able to pass some parameters to this API endpoint such as the hotel name, the number of rooms and the postcode, so we can use this API to create different hotel. These parameters can be sent in the body of the request. To access the parameters in the body of the request with Express.JS, we need to add the module `body-parser` to the `cyf-hotels-api` project:
-
-```
-npm install --save body-parser
-```
-
-Then include it in the `server.js`:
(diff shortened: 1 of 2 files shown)


docs/db/week-3/lesson.md
```diff
@@ -80,19 +80,12 @@ In the following, we will add a new API endpoint to create a new hotel in the ta
...
 ```
 
-As we create a new record in the database, we will add a new POST endpoint in the `cyf-hotels-api` project from last class. Moreover, we need to be able to pass some parameters to this API endpoint such as the hotel name, the number of rooms and the postcode, so we can use this API to create different hotel. These parameters can be sent in the body of the request. To access the parameters in the body of the request with Express.JS, we need to add the module `body-parser` to the `cyf-hotels-api` project:
-
-```
-npm install --save body-parser
-```
-
-Then include it in the `server.js`:
+As we create a new record in the database, we will add a new POST endpoint in the `cyf-hotels-api` project from last class. Moreover, we need to be able to pass some parameters to this API endpoint such as the hotel name, the number of rooms and the postcode, so we can use this API to create a different hotel. These parameters can be sent in the body of the request. To access the parameters in the body of the request with Express.JS, we need add the following line in the `server.js` file of the `cyf-hotels-api` project:
```
illicitonion, line 83: Should this include a direction about _where_ in the file to include it? e.g. something like "After the define the `app` variable" or "Before we declare the first handler" or "before the call to `app.listen`" or similar?

illicitonion (approved):
Looks great to me! I left one question

docs/db/week-3/lesson.md
MarkFarmiloe: Yes, I wondered about this, but I don't think it matters, as long as it's after app is instantiated, it's just another bit of configuration that is only used once the app.listen is called.
At least that what I believe, I'm not at all certain, so hopefully someone else can confirm one way or other.
Or I could test my theory of course 🤦

PR #266, opened 2021-05-23
Add testing primer ( with Jest )

Dedekind561:
## What does this change?

Module: Guide - Jest intro
Week(s): n/a

## Description

Adds a guide to understanding tests written using Jest.
Currently a WIP

@ChrisOwen101 

-----
[View rendered docs/guides/intro-to-tests.md](https://github.com/CodeYourFuture/syllabus/blob/add-jest-intro/docs/guides/intro-to-tests.md)


docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+- Understanding test feedback ( using Jest )
+
+## Code specification
+
+`createWordCount` 🧮
+
+Let's consider a function called `createWordCount` that needs to work as follows:
+
+- it takes one argument - a _string_ of words separated by spaces, e.g. `"hello world"`
+- returns an _object_ with a count of each word in the string
```
illicitonion: We're using this for JS1 week 1, so objects aren't something our trainees have encountered yet - maybe let's rework this example to just return a count of words (i.e. `createWordCount("hello world") === 2`)?

There are still plenty of edge-cases we can bring up (e.g. empty string, do numbers count as words, hyphenated words). Speaking of edge-cases, can we explicitly call these out? Something like:

> When thinking about a specification, and what tests you'll need, you should be thinking _both_ about how people would _normally_ use the function, and about _unusual_ examples that may be confusing or surprising (often called "edge cases"). For example, with our `createWordCount` example, what should happen when the empty string is used as an argument? What should happen for hyphenated words (e.g. North-West)? Do numbers count as words? What should `createWordCount("There were 17 trainees in the North-West class")` return?
>
> You should make sure you think about these edge-cases when you're approaching a problem, to make sure you solve it completely.
>
> Often when you're given a problem, not all of these edge-cases are described for you, so you'll need to use your best judgement (this is true as a professional developer too!). Writing down what you decided (either in a specification, or as a test) can be useful to show other developers what you decided. It will help them to understand your code, and also means that if they have different ideas about some of the cases, you can have a discussion about it. So in this example, we may write a test showing that numbers are counted as words (or that numbers are not counted as words, depending on what we decided!).

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+
+```js
+const input = "hello";
+const actualResult = createWordCount(input);
+const expectedResult = {
+  hello: 1,
+};
+```
+
+In this section, we're declaring some important values that we will need to compare later on in the test.
```
illicitonion: I can see some of our trainees reading this as a required recipe (you need a variable named `input` and a variable named `actualResult`, etc) - while the note after is helpful, it may also be worth adding an introductory sentence/paragraph before this saying something like "We can write any code we want in the test function. The important thing is that we have at least one assertion, but we can use all of the techniques we learn in javascript (like using variables, or calling functions) to help us to write our test."

Which actually makes me thing... Maybe we can build this up the other way around? Start with a test that looks like:

```js
test("will count the number of words for a single word string", function () {
    expect(createWordCount("hello")).toEqual({hello: 1});
});
```

so that we can quickly explain assertions, and then break apart the test to use variables and show that it can make it more clear?

illicitonion (commented):
I like the structure of building from specification to cases to tests, then breaking down a test, then interpreting feedback :) Thanks for putting this together!

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+In this introduction to testing we'll cover the following things:
+
+- Code specification
+- What is a test ?
+- Anatomy of a test ( using Jest )
+- Understanding test feedback ( using Jest )
+
+## Code specification
+
+`createWordCount` 🧮
```
alismith-dev: I find the "create" naming here a bit confusing. In my head, this function is performing a calculation, more than "creating" something. The object is just the data format for the results.

I wonder if `countWords` would be more intuitive?

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+}
+```
+
+## 🧪 What is a test ?
+
+We can now define what a test is in software engineering.
+
+> 🔑 A **test** is any piece of code that **checks a certain unit of code** ( typically a function ) **behaves in a particular way**.
+
+For example, we could write a test to check that `createWordCount` behaves in a particular way when passed a string containing a single word. For this example, we're using a popular JavaScript testing framework called [Jest](https://jestjs.io/docs/getting-started) to write the test code. We'll examine the example below written using Jest, and then break down the different parts of the test code to understand how it all works.
```
alismith-dev: This is a bit of a nitpick, but

```suggestion
For example, we could write a test to check that `createWordCount` behaves in the way we described above when passed a string containing a single word. For this example, we're using a popular JavaScript testing framework called [Jest](https://jestjs.io/docs/getting-started) to write the test code. We'll examine the example below written using Jest, and then break down the different parts of the test code to understand how it all works.
```

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Anatomy of a test
+
+#### 1. Test description
+
+```js
+test("will count the number of words in a single word string", function () {});
+```
+
+We use a function in Jest called `test` to write a test description - `"will count the number of words for a single word string"` which describes the behaviour we expect in this particular test. There is also a function `function () {}` in which we will write our test code.
```
alismith-dev: I wonder if it's worth calling out that `test` is injected by Jest?

I think I saw in another PR (Daniel's maybe?) that we were explicitly importing the variable - which I assume was discussed at the syllabus team meeting that I missed - perhaps it is worth discussing that here?

docs/guides/intro-to-tests.md
alismith-dev: In my experience, extracting variables out of a complex example can sometimes lose trainees when they're not confident with expressions/statements. On the other hand, we hopefully should have addressed that so maybe it's not a concern.

I do agree that showing a inlined example is probably worth doing. It's common enough practice that trainees looking for more materials online will see it, so I think it's worth showing and discussing as part of the class.

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### 2. Test setup
+
+```js
+const input = "hello";
+const actualResult = createWordCount(input);
+const expectedResult = {
+  hello: 1,
+};
```
alismith-dev: What do you think about moving `expectedResult` down to the assertion section below? This way, we roughly follow the Arrange-Act-Assert pattern (although we're not really doing much "arranging").

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+In this section, we're declaring some important values that we will need to compare later on in the test.
+We declare an `input` variable that stores a value that we will pass to the function we are testing - `createWordCount`.
+We declare an `actualResult` variable that stores the **return value** of our function `createWordCount`.
+We declare an `expectedResult` which stores what we **expect** the `actualResult` to be when our function is behaving according to the test description.
+
+❗ Note: The test setup will vary from test to test - so you can expect to see different variable names altogether in different tests. The setup here is done in order to make it clearer what values will be compared later on in the test.
```
alismith-dev: Fwiw, if you really want to call attention to something, we can use Docusaurus' "admonitions" feature: https://docusaurus.io/docs/markdown-features/admonitions

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+
+The function [`toEqual`](https://jestjs.io/docs/expect#toequalvalue) ( which is part of the Jest framework ) is used to check that `actualResult` ( `createWordCount`'s return value ) and `expectedResult` are equal to each other. There are [many other functions](https://jestjs.io/docs/using-matchers) like `toEqual` which we can use to make different assertions.
+
+### Pass ✅ or Fail ❌
+
+Finally, our test can either pass or fail.
+
+- If `actualResult` is equal to `expectedResult` then the test has passed ✅.
+- If `actualResult` is _not_ equal to `expectedResult` then the test has failed ❌.
+
```
alismith-dev, line 229: I think we should add something here along the lines of:

> **Don't panic if a test is failing!**
> Remember: tests are here to help you solve the problem! If you carefully read the test feedback it will tell you what the problem is.

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+![Failing test case for createWordCount](./assets/test-fails-case.png)
+
+Jest aims to give feedback that makes it simple to see why test has failed. It shows us several things:
+
+- the line number for the failing test (line 9)
+- shows us `expected` - which corresponds to our `expectedResult`
+- shows us `received` - which corresponds to our `actualResult`
+
+We can see the test feedback is telling us that `received` is `undefined`. In other words, `actualResult` ( `createWordCount`'s return value) is `undefined`.
+
```
alismith-dev, line 256: Again, apologies for missing the syllabus team call when this I suspect this was discussed, but what are we doing about "CI" (if anything)? I.e. if we're using Github Actions, should we also show how to read this output?

alismith-dev (commented):
Thanks for getting this started @Dedekind561! I've left a few comments, but I like the overall approach 🙂 

Is this designed to be run as coursework? Or guided by a teacher?

docs/guides/intro-to-tests.md
Dedekind561: I agree - it did occur to me that they won't have seen objects but I ploughed on anyway. I think the simplified version will work a lot better.

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+## What does this cover
+
+This introduction is designed to give you a high-level overview on the basics of testing, which you'll need when you're tackling the problems in your coursework at Code Your Future.
+In this introduction to testing we'll cover the following things:
+
+- Code specification
+- What is a test ?
+- Anatomy of a test ( using Jest )
+- Understanding test feedback ( using Jest )
+
```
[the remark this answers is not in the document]
Dedekind561: I did have this in! - but then stripped it out because it was getting too technical. I think your points are far more concise and to the point though :)

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+In this case, we would expect the `output` to be:
+
+```js
+{
+    "north" : 2,
+    "by" : 1,
+    "west" : 1
+}
+```
+
```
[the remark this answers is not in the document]
Dedekind561: Absolutely :+1:

docs/guides/intro-to-tests.md
Dedekind561: :+1: Agreed - I think we should stick to "count"

docs/guides/intro-to-tests.md
Dedekind561: To be fair I'm not sure what the situation is with injected globals vs required variables - I'll find out at the syllabus meeting.

docs/guides/intro-to-tests.md
Dedekind561: Ohh nice didn't know you could do that

docs/guides/intro-to-tests.md
Dedekind561: Excellent idea - love it :muscle:

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+![Failing test case for createWordCount](./assets/test-fails-case.png)
+
+Jest aims to give feedback that makes it simple to see why test has failed. It shows us several things:
+
+- the line number for the failing test (line 9)
+- shows us `expected` - which corresponds to our `expectedResult`
+- shows us `received` - which corresponds to our `actualResult`
+
+We can see the test feedback is telling us that `received` is `undefined`. In other words, `actualResult` ( `createWordCount`'s return value) is `undefined`.
```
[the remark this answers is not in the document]
Dedekind561: Yeah I think so

docs/guides/intro-to-tests.md
Dedekind561: Not sure myself tbh - will find out at syllabus tonight I think.

docs/guides/intro-to-tests.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Further reading
```
ChrisOwen101, line 257: Perhaps adding some suggested reading/watching would be good here?

I always enjoy Coding Train: https://www.youtube.com/watch?v=S3QwafQEvSs

ChrisOwen101 (approved):

PR #274, opened 2021-06-03
Switch to teaching functional set state

### LINKED ISSUE #224: React Week 2: setting state based on previous state not taught correctly
The syllabus consistently uses examples that change state based on previous state that are not correct (according to my understanding). There are multiple, but the first block of code with this issue is below:

```js
function Counter() {
  const [count, setCount] = useState(0);

  function incrementCount() {
    setCount(count + 1); // this line is wrong, it should be setCount(count => count + 1);
  }

  return (
    <div>
      <button onClick={incrementCount}>Click me</button>
      <p>You clicked {count} times</p>
    </div>
  );
}
```

**What is wrong with this content?**

I have created [some example code](https://codesandbox.io/s/broken-state-changes-88em7?file=/src/index.js) illustrating why this approach doesn't work, and you can also [read the react docs](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous) for the official recommendation.

**How could this be fixed?**

Replace all instances where state is updated based on previous state to use arrow functions instead of references to locally scoped variables. Add a paragraph to make it clear to students that arrow functions should always be used when changing state based on previous state.

**Link to content**

https://syllabus.codeyourfuture.io/react/week-2/lesson/

alismith-dev:
## What does this change?

Module: React
Week(s): 2

## Description

Closes https://github.com/CodeYourFuture/syllabus/issues/224.

As discussed in the above ticket, updating state based on previous state can have some gotchas - specifically around calculating the new state based on previous state. So this PR changes how setting state is taught. We now:

1. Teach set state using a hard-coded value (`setCount(1)`)
2. Demonstrate the gotchas through an example (multiple set states)
3. Teach set state using a functional update (`setCount((c) => c + 1)`)

I originally pushed back on doing this because I was concerned about the additional complexity, however I've since come around and now think this is the way forward. Partly this is based on my recent experience with teaching London Class 7, where I experimented with demonstrating the problem and then teaching functional set state. I found that the trainees were fairly responsive to this.

I also think that it reinforces the point made during the lesson about set state needing to notify React that data has changed and a re-render is required. Ultimately, this is the underlying source of the gotcha so I think it's useful to dive into it a bit more.

@chrisdrifte opened the ticket originally. Thanks for the useful CodeSandbox - I really liked the idea around using multiple set states to demonstrate the issue, I think it's quite illustrative. Feedback welcome on what I've written here!

## Rendered Pages

-----
[View rendered docs/react/week-2/lesson.md](https://github.com/CodeYourFuture/syllabus/blob/react-functional-set-state/docs/react/week-2/lesson.md)

### DIFF
--- a/docs/react/week-2/lesson.md
+++ b/docs/react/week-2/lesson.md
@@ -317,14 +317,14 @@ const [unreadMessages, setUnreadMessages] = useState(5);
 
 ### Updating State
 
-Our Counter isn't very useful right now! Let's make it more useful by getting `count` to actually count up ([interactive example](https://codesandbox.io/s/usestate-counter-lorv5?file=/src/Counter.js)):
+Our Counter isn't very useful right now! Let's make it more useful by getting `count` to actually count up ([interactive example](https://codesandbox.io/s/usestate-setting-state-lorv5?file=/src/Counter.js)):
 
 ```js
 function Counter() {
   const [count, setCount] = useState(0);
 
   function incrementCount() {
-    setCount(count + 1);
+    setCount(1);
   }
 
   return (
@@ -342,11 +342,11 @@ Our component now has a `<button>`, which will call the `incrementCount` functio
 <button onClick={incrementCount}>Click me</button>
 ```
 
-The `incrementCount` function then calculates the **new** state by adding 1 onto the current `count`. And then calls `setCount` to set the new state:
+The `incrementCount` function calls `setCount` to set the new state to be `1`:
 
 ```js
 function incrementCount() {
-  setCount(count + 1);
+  setCount(1);
 }
 ```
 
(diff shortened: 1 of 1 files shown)


40thieves:
⚠️ Rebased on `main` and force pushed

docs/react/week-2/lesson.md
```diff
@@ -364,7 +364,53 @@ function Counter() {
...
+  setCount((c1) => {
+    return c1 + 1;
+  });
+  setCount((c2) => {
+    return c2 + 1;
+  });
+}
+```
+
+When we pass a callback function to `setCount`, React will pass us the **latest version** of the count state. In this example, the variables are named `c1` and `c2` to make it clear that they are different from the `count` variable. Now if we add `console.log(c1)` and `console.log(c2)` then we can see that they get the correct values.
```
illicitonion, line 409:
```suggestion
When we pass a callback function to `setCount`, React will pass us the **latest version** of the count state. In this example, the variables are named `c1` and `c2` to make it clear that they are different from the `count` variable. Now if we add `console.log(c1)` and `console.log(c2)` then we can see that they get the correct values.

Reminder: `c1` and `c2` are just arbitrary names we pick to be the name of the parameter to the function we're passing - we could have called them both `c`, and everything would still work - they're not related to each other, as they are defined in different scopes.
```

As they tend to struggle with scope, I'm a little concerned that without this clarification, trainees may think that these need to be different, or have some relation to each other.

illicitonion (approved):
Looks good :)

docs/react/week-2/lesson.md
alismith-dev: Yeah, good suggestion thanks.

PR #313, opened 2021-08-04
Mention Git/Terminal courseworks on Coursework pages

### LINKED ISSUE #310: Clarify when the Git coursework should be assigned
The git classes are assigned in the background - we should make it a lot more clear when they should be done, and highlight it in those weeks' homeworks.

See also https://codeyourfuture.slack.com/archives/C01DANTAGRK/p1627729083001800 from @ClaireBickley

illicitonion:
We already had this for JS2W1 but apparently missed the others.

Fixes #310


alismith-dev (approved):
LGTM :shipit:

PR #324, opened 2021-09-26
Add draft module systems lesson

### LINKED ISSUE #218: Add require, file paths & NPM to JS-3 Week 3
Module(s): JS3
Week(s): 3

Add lesson content to show how to use the `require` function and packages from NPM. This would also include some content about file paths (including relative paths with `./`).

We could consider reworking some of the content in https://github.com/CodeYourFuture/syllabus/pull/189, as it includes interaction with a simple (and fun!) NPM package, which trainees could then recreate to learn about `require`.

There have been some recent discussions about when we teach these important aspects of JS.

In this thread: https://github.com/CodeYourFuture/syllabus/pull/166#issuecomment-770410775, it was pointed out that we don't teach anything about the `require` function (and `import`) or interact with NPM packages until the React module.

Additionally, in this thread: https://github.com/CodeYourFuture/syllabus/pull/166#issuecomment-770410775, we discussed how file paths in the context of JS are often forgotten about, but are quite important in modern apps. Similar to the point above, currently we do not teach about file paths (in JS) until the React module.

There are 2 problems with this:

1. Paths, `require`/`import` and NPM are very useful outside the React module, so we should teach them separately
2. The first week of React is already fairly packed with content, so there is not much time to practice the concepts

**Additional context**

This was discussed at the syllabus team meeting on 2021-03-17.

**Who might need to know about this change?**

@ChrisOwen101 
@SallyMcGrath 
@illicitonion

alismith-dev:
## What does this change?

Module: Probably JS3, but see discussion below
Week(s): Again see discussion below

## Description

Closes https://github.com/CodeYourFuture/syllabus/issues/218.

Adds a new lesson (or lesson section) about module systems, file paths and NPM.

#### Placement

I sort of wrote this in a vacuum and wasn't quite sure where to put it in the end. So for now, I've put it in a new folder in JS3. I figured we can decide where to put it as part of the discussion. My thoughts on placement:

- The file paths section is probably earlier in the course, to go along with the terminal lessons. However those are designed as at-home coursework, and I wrote this as a delivered lesson so I don't think it could be extracted as-is. I can probably re-work it to be more suited to a coursework exercise
- The other sections probably fit into JS3, as suggested in https://github.com/CodeYourFuture/syllabus/issues/218. Looking at the current content, week 1 looks the lightest in terms of taught content so probably can go in there. However I haven't taught JS3 in it's current form so I don't know which week it would work best in

#### Exercises

I was trying to keep it relatively short, so there's only a handful of exercises. Should there be some more?

There's also no exercises which involve `require`ing a "regular" file (as opposed to an NPM package). Perhaps we should add one? 

I somewhat cannibalised the content of https://github.com/CodeYourFuture/JavaScript-Core-1-Challenges/pull/25. However this expands the scope of the cowsay exercise a bit (touches on `npx`, and reading the docs to some extent). We could include some more of this in the lesson?

#### Note for teachers

I picked to teach CJS first as it's more likely to Just Work™️ for trainees, especially as we've generally run small JS examples in Node up to this point.

#### Open questions

- Should we introduce other file path conventions (`~`)?
- Should we introduce absolute paths? (Or just introduce the relative path terminology?)
- Should we touch on module (file) naming/structuring?
- Should we tell trainees not to install packages unless we tell them to? So that they don't "cheat" on exercises

@CodeYourFuture/syllabus-team


docs/js-core-3/module-systems/lesson.md
```diff
@@ -0,0 +1,399 @@
...
+
+There are a **huge** number of modules written by other programmers (usually called _packages_) available for us to download and use in our applications. This gives us superpowers to solve complex problems more quickly, since many common problems have already been solved for us. We'll look later at how we can do this in our own applications.
+
+## How does a module system work?
+
+As we've just seen, a module system helps us to split our code into smaller modules.
+
+:::tip
+In nearly every module system, a "module" is really just another name for a file. So we'll just call a "module" a "file" from now on.
+:::
```
DomVinyard, line 44: I always was struck by Bret Victor's bit about how 'file' is quite a poor abstraction. How the 'fundamental units' of computation are expression, line, function, class, module, etc. Wheras a file is just an implementation detail of the underlying hardware.

Not to say this is the best way to teach it (everyone understands what a file is) thought just an interesting aside.

docs/js-core-3/module-systems/lesson.md
```diff
@@ -0,0 +1,399 @@
...
+    ├── coursework
+    │   ├── exericse1.js
+    │   └── exercise2.js
+    └── notes
+        └── js3
+            └── week1.txt
+```
+
+:::info
+This style of "drawing" folders comes from a command line program called `tree`. The name refers to the idea that your file system is structured like a bit like a family tree.
```
DomVinyard, line 103: _family_ tree might be the a slightly confusing metaphor. Implies two parents to each child?

docs/js-core-3/module-systems/lesson.md
```diff
@@ -0,0 +1,399 @@
...
+
+You might start to recognise the file path patterns! You've actually already used file paths when using the `git add` command, although you might not have realised it. File paths are used in many other commands too.
+
+As we looked at earlier, we can also use file paths when importing files in JavaScript. We can now explain what this code does:
+
+```js
+const findWidget = require("./widget-finder");
+```
+
+Now we know that the `./widget-finder` file path refers to a file in the same folder called `widget-finder.js`. We can write out the `.js` bit if we want, but if you leave it off, then the module system will assume that you mean `.js`.
```
DomVinyard, line 182: maybe we should leave the `.js` in the examples (for clarity) - and then teach that it can actually be dropped a little later, once it's been cemented

docs/js-core-3/module-systems/lesson.md
```diff
@@ -0,0 +1,399 @@
...
+
+const findWidget = widgetFinderExports.findWidget;
+const findAllWidgets = widgetFinderExports.findAllWidgets;
+```
+
+## NPM
+
+_Node Package Manager_ - or NPM for short - is a massive collection of JavaScript code that you can download and use in your apps (for free!). This is extremely helpful for us, as we can concentrate on the specific problem we want to solve, which helps build our apps faster.
+
+As the name implies, NPM is a tool for managing _packages_. So what is a package? Packages aren't anything special really! They're made up of JavaScript files, just like any other file that you've written. A developer has then uploaded those files into the NPM collection, allowing any one to dowload them and use them for free.
```
DomVinyard, line 233: > So what is a package

Doesn't really go on to answer the question. Perhaps:

> ...Packages aren't anything special really! **A package is just a collection of files that helps you to add some functionality to your application**. A developer has then...

taken from https://cleverbeagle.com/blog/articles/what-are-javascript-packages-and-dependencies

docs/js-core-3/module-systems/lesson.md
```diff
@@ -0,0 +1,399 @@
...
+| 3. Add `const figlet = require("figlet");` at the top of the file.                                              |
+| 4. Run `npm run ascii`.                                                                                         |
+
+## Using ECMAScript Modules
+
+In some code that you see on the internet, you may find a different module system being used. Don't panic! It is very similar to the CommonJS system that you just learned about.
+
+This new module system is typically called ECMAScript Modules, or ESM for short.
+
+Why are there 2 module systems? Unlike `require` & `module.exports`, which use bits of JS that use already know (functions, variables and objects), the ESM system introduces some new syntax. ESM is built into the language itself.
```
DomVinyard, line 318: Maybe instead of referring to them primarily as Commonjs vs ESM.. we should refer to them as 'require style' vs 'import style' and mention the official names just as a parenthetical aside - is that a stupid idea, just thinking out loud here -

docs/js-core-3/module-systems/lesson.md
```diff
@@ -0,0 +1,399 @@
...
+
+We now have access to both the `findWidget` and `findAllWidget` functions.
+
+:::tip
+You can think of this `import` syntax like destructuring an object.You can imagine that we are importing an object from the `widget-finder` file, then destructuring the exported references out of that object.
+:::
+
+## Using module systems in your code
+
+Unfortunately, deciding which module system to use is currently quite complicated. Both systems are actively being used in different apps, but often they are **incompatible** with each other 😕
```
DomVinyard, line 391: Seems quite a negative intro. Makes it seem like a massive issue when in practice it doesn't actually cause a headache allll that often (?). Maybe we should start this section with the tip at the bottom as the main introduction paragraph, with the disclaimer that it's occasionally a pain once the distinction between node and browser is made clear

DomVinyard (commented):
Awesome 🤩 - great content, threw a few comments in but nothing major at all, just some thinking out loud 🙂

docs/js-core-3/module-systems/lesson.md
```diff
@@ -0,0 +1,399 @@
...
+./index.html
+```
+
+Let's break this down into pieces:
+
+- We start off with a dot character (`.`). This has a special meaning in file paths. It is a shortcut that means the folder of the starting location. So in this example, it means the JS3 `project` folder
+- Next there's a slash character (`/`). This separates the current folder that we just saw from next piece, the `index.html` file
+- Finally we have the name of the destination location, in this case `index.html`
+
+![File path to index.html. An arrow goes up from script.js to project, then down to the index.html file](./assets/index-html-file-path.png)
```
gregdyke, line 123: I think this image actually incorrectly shows ../index.html

docs/js-core-3/module-systems/lesson.md
```diff
@@ -0,0 +1,399 @@
...
+## Using module systems in your code
+
+Unfortunately, deciding which module system to use is currently quite complicated. Both systems are actively being used in different apps, but often they are **incompatible** with each other 😕
+
+The main reason for this is that ESM was created after CJS and so a lot of older code has not been moved away from CJS. To make things worse, in some cases, it is currently impossible to move code to ESM.
+
+For example, when running code in Node, then CJS will tend to work much better than ESM. Similarly when running code in a browser, then CJS will not work at all, but ESM will work.
+
+:::tip
+The rule of thumb is when your code will run in Node then you should use the CJS module system. And when your code will run in the browser, use the ESM module system.
```
gregdyke, line 398: what about react? (and where does that leave us with helping students with TDD, especially if we'd like them to spend more time in the browser?)

gregdyke (changes requested):
I think several parts should be broken out: navigating file systems is it's own thing that might come in useful at various parts of the syllabus, but doesn't need repeating to make this longer (also I think that part is very good). Possibly also the npmish part?

If we're assuming coming in at the end of js3, it might be useful to show how the tv show project might be broken down as it's becoming increasingly unwieldy.

Is it worth mentioning that if you have complete separation of concerns, you can just have 2 js files in 2 script tags?

40thieves:
This has been stalled for quite some time so will close. I think we may address this as part of an overall JS rethink.

Issue #325, opened 2021-09-30
SQL Module Preparation - DB name is spelled differently


ghost:
On the Preparation page for the SQL module the database create is cyf_hotel but during the lessons it is cyf_hotels, obviously if you just copy the code it then doesn't work because that database doesn't exist.

PR #326, opened 2021-09-30
Update prep.md

ghost:
## What does this change?

#325 
SQL Module Preparation - DB name is spelled differently

## Description
On the Preparation page for the SQL module the database create is cyf_hotel but during the lessons it is cyf_hotels, obviously if you just copy the code it then doesn't work because that database doesn't  exist.

@SallyMcGrath 

## Rendered Pages

-----
[View rendered docs/db/prep.md](https://github.com/ochthenoodle/syllabus/blob/patch-1/docs/db/prep.md)

### DIFF
--- a/docs/db/prep.md
+++ b/docs/db/prep.md
@@ -68,22 +68,22 @@ Remember the new username and password details - you will need them later.
 
 ### Creating a new database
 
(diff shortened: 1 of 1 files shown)


SallyMcGrath:
Thanks! @ochthenoodle

PR #354, opened 2021-12-12
Update navigation to pass tests

### LINKED ISSUE #296: Archive MongoDB Module
Module(s): MongoDB
Week(s): all

As per Syllabus team meeting, archive Mongo DB for now and focus on SQL module. 

- [ ] collate and then archive Mongo issues
- [ ] send any issues useful for SQL to SQL Garden
- [ ] take MongoDB links off Syllabus website?
- [ ] let mongo channel know

See team minutes -  extensive discussions -  this action was agreed 23/06/21

**Who might need to know about this change?**

@gregdyke

SallyMcGrath:
The latest Docusaurus won't run with improper config so I've updated the navigation to pass the tests.

I've also removed Mongo from the main nav, but not the module from syllabus. 

## What does this change?

Module: MongoDB
Week(s): n/a

## Description

Removed all the 'left' stuff so the navigation would build without these errors:

```
A validation error occurred.
The validation system was added recently to Docusaurus as an attempt to avoid user configuration errors.
We may have made some mistakes.
If you think your configuration is valid and should keep working, please open a bug report.
```

ValidationError: "navbar.items[1].items[0].position" is not allowed

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -49,204 +49,156 @@ module.exports = {
         {
           to: "Overview",
           label: "Overview",
-          position: "left",
+          
         },
         {
+          to: "Overview",
           label: "Modules",
-          position: "left",
+          
           items: [
             {
               href: "https://introtocoding.codeyourfuture.io",
               label: "Intro To Coding",
-              position: "left",
-              className: "outside-link",
+              
             },
             {
               href: "https://fundamentals.codeyourfuture.io/",
               label: "Fundamentals",
-              position: "left",
-              className: "outside-link",
             },
             {
               href: "https://personaldevelopment.codeyourfuture.io/em-modules/induction-week",
               label: "Induction",
-              position: "left",
-              className: "outside-link",
             },
             {
               to: "git/overview",
               label: "Git and Github",
-              position: "left",
             },
             {
-              to: "html-css/index",
+              to: "html-css/",
               label: "HTML-CSS",
-              position: "left",
             },
             {
-              to: "js-core-1/index",
+              to: "js-core-1/",
               label: "JavaScript Core 1",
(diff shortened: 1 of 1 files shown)


docusaurus.config.js
```diff
@@ -49,204 +49,156 @@ module.exports = {
...
+          to: "Overview",
           label: "Modules",
-          position: "left",
+          
           items: [
             {
               href: "https://introtocoding.codeyourfuture.io",
               label: "Intro To Coding",
-              position: "left",
-              className: "outside-link",
```
ChrisOwen101, line 62: image

These classNames add the nice little "this link opens another website" styling to the buttons. It's nice to have but if ti means the config won't compile then I guess it'll have to go

docusaurus.config.js
```diff
@@ -49,204 +49,156 @@ module.exports = {
...
+              
             },
             {
               href: "https://fundamentals.codeyourfuture.io/",
               label: "Fundamentals",
-              position: "left",
-              className: "outside-link",
```
ChrisOwen101, line 68: Same as above

docusaurus.config.js
```diff
@@ -49,204 +49,156 @@ module.exports = {
...
             },
             {
               href: "https://personaldevelopment.codeyourfuture.io/em-modules/induction-week",
               label: "Induction",
-              position: "left",
-              className: "outside-link",
```
ChrisOwen101, line 74: Same as above

ChrisOwen101 (approved):

docusaurus.config.js
SallyMcGrath: It does it anyway now, for some reason.

PR #373, opened 2022-02-01
Fix code examples in SQL module lessons

szemate:
## What does this change?

Module: SQL
Week(s): 2 and 3

## Description

Fix some errors in the Node.js code examples in the SQL module lessons.
- Fix error handling
- Fix promise chaining
- Resolve inconsistency between weeks 2 and 3
- Add syntax highlighting

See the individual commit messages for details.

## Rendered Pages

-----
[View rendered docs/db/week-2/lesson.md](https://github.com/szemate/syllabus/blob/db/fix-code-examples/docs/db/week-2/lesson.md)
[View rendered docs/db/week-3/lesson.md](https://github.com/szemate/syllabus/blob/db/fix-code-examples/docs/db/week-3/lesson.md)

### DIFF
--- a/docs/db/week-2/lesson.md
+++ b/docs/db/week-2/lesson.md
@@ -309,6 +309,20 @@ In the code above:
 - We create a new pool where we specify the credentials to connect to the cyf_hotels database
 - We then create a new /hotels endpoint where we use the method `query()` to send a SQL query to load all the hotels from the table `hotels` and return the results with `result.rows`. You can write any valid SQL query that you learned in the `query()` method!
 
+In the example above the second argument to `pool.query` is a _callback function_. This callback function is called by `pool.query` as soon as the database server returns the result.
+
+There is another way of sending queries using _promises_. This is very similar to how the `fetch` API works in the browser. If we don't specify a second argument to `pool.query`, it will return a promise, and the request handler above will look like:
+
+```
+app.get("/hotels", function(req, res) {
+    pool.query('SELECT * FROM hotels')
+        .then((result) => res.json(result.rows))
+        .catch((error) => {
+            console.error(error);
+            res.status(500).json(error);
+        });
+```
+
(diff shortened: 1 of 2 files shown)


gregdyke (commented):
Nice! I'm glad we're moving towards more consistency with this. I'm tempted to approve and we can improve even further later.

But I think this would be a good time to really tighten up how we deal with promises and callbacks:

 - I think we should either have callbacks or promises, not both
 - I think we won't ever properly teach promises, so teaching promise chaining is an iffy practice - I would be tempted to leave the ugly embedding
 - Alternatively, we could teach await. I think we should do that in the DB module because it's easier to reason about.

So my suggestion: 
 
 - get rid of the callback syntax entirely
 - leave the multiple embedded promises for now

Then a followup would be to consider whether teaching await/async (and associated try/catch, or associated "catch exceptions in a middleware") is worth it

gregdyke:
although in fairness, we already have implicit promise chaining in the client side without realising it when we do

`fetch('/api/bookings').then((res) => res.json()).then(setBookings)`

So maybe it's not that bad?

szemate:
> get rid of the callback syntax entirely

That's a good point, I will do that.

> we already have implicit promise chaining in the client side without realising it when we do

I think promise chaining is explicitly taught at https://syllabus.codeyourfuture.io/js-core-3/week-2/lesson#using-fetch. The example included is:

```js
getMilkFromShops
  .then((milk) => {
    console.log(`I've got the milk`);
    milk.drink();
    return milk.bottle;
  })
  .then((bottle) => {
    console.log(`I'm going to recycle the bottle`);
    bottle.recycle();
  })
  .catch((error) => console.warn("Oh no, I dropped the milk"));
```

gregdyke:
Oh. I meant promise chaining where the returned object is a promise is. But if we already do promise chaining, I'm sure we can explain "you can return a value or a promise here and then chain with then after" quite easily :)

SallyMcGrath:
@all-contributors please add @szemate for content and code

SallyMcGrath:
@all-contributors please add @MarkFarmiloe  for content and code

PR #375, opened 2022-02-04
Update lesson.md

jcholyhead:
- 

## What does this change?
Module: SQL
Week(s): 3

## Description

Parameterised user input in reading data example. Previous example allowed SQL injection through the API query parameter.

Syllabus team
## Rendered Pages

-----
[View rendered docs/db/week-3/lesson.md](https://github.com/jcholyhead/syllabus/blob/patch-1/docs/db/week-3/lesson.md)

### DIFF
--- a/docs/db/week-3/lesson.md
+++ b/docs/db/week-3/lesson.md
@@ -194,13 +194,14 @@ Another functionality which could be useful is to filter the hotel with a keywor
 app.get("/hotels", function (req, res) {
   const hotelNameQuery = req.query.name;
   let query = `SELECT * FROM hotels ORDER BY name`;
-
+  let params = [];
   if (hotelNameQuery) {
-    query = `SELECT * FROM hotels WHERE name LIKE '%${hotelNameQuery}%' ORDER BY name`;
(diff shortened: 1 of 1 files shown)


szemate:
While this change makes sense, I don't think we should put code like this into the examples without explanation. Anyway, this is not the only SQL command that allows SQL injection, almost all the code examples in the lesson do.

jcholyhead:
> While this change makes sense, I don't think we should put code like this into the examples without explanation. Anyway, this is not the only SQL command that allows SQL injection, almost all the code examples in the lesson do.

I think this is the only string concatenation example in the lesson file (I haven't spotted any others). All the others use parameterised queries. Parameterising a LIKE clause is a little messy because of the wildcards, which is why I guess it was left in as a string concatenation.

szemate:
> All the others use parameterised queries.

Yes you're right. I think it's good to merge then.

szemate:
If no-one from the syllabus team has any objection, I will merge it today before 10 so that we have the fixed syllabus for today's class.

SallyMcGrath:
@all-contributors please add @jcholyhead  for content and code

PR #377, opened 2022-02-16
Create exercise1.js

calarasiu:
## What does this change?

Module:3
Week(s):1

## Description

As a class we should step through solving all of the issues in this small piece of code.  Be sure to use The Debugging Framework that we discussed earlier


SallyMcGrath:
Please read the contributing guidelines and come to me if you have questions. Thanks! 🙏 ❤️

Issue #428, opened 2022-04-30
Technical syllabus links are down 


zero-point:
**Describe the bug**

I'm not sure this is the right place to report it, but the links to the technical content on the syllabus don't work. E.g. https://syllabus.codeyourfuture.io/html-css/index

**What is wrong with this content?**
Links not working 

**How could this be fixed?**
Update links I guess 

**Link to content**
https://syllabus.codeyourfuture.io/

**Additional context**

**Who else might need to know about this?**

SallyMcGrath:
Hi hi Irina - can you tell me where this link is? (When I choose from the menu it works.) I found a broken link on https://syllabus.codeyourfuture.io/Overview -- is that where you where?

zero-point:
Accessing the links from the Overview page https://syllabus.codeyourfuture.io/Overview

SallyMcGrath:
fixed by #429

PR #429, opened 2022-05-03
fix bad links /index

SallyMcGrath:
## What does this change?

Module: OVERVIEW
Week(s):

## Description

this page had bad links because it was written for a previous version of docusaurus

 @zero-point


SallyMcGrath:
Fixes #428

PR #463, opened 2022-07-28
Remove "Structuring Data" from Lesson 2 summary

IanTaite:
There's nothing in lesson 2 related to "structuring data" as far as I can tell, so I've removed the "Structuring Data" label from Lesson 2.

## What does this change?

Module: HTML/CSS
Week(s): 2

## Description

Lesson 2's description is "Forms and Structuring Data", but there's nothing in the lesson content about structuring data.

@stevoland @jamesgorrie @frosas @skinofstars 

## Rendered Pages

-----
[View rendered docs/html-css/index.md](https://github.com/CodeYourFuture/syllabus/blob/IanTaite-patch-2/docs/html-css/index.md)

### DIFF
--- a/docs/html-css/index.md
+++ b/docs/html-css/index.md
@@ -7,7 +7,7 @@ sidebar_label: Overview
 | Weeks                          | Content                           | Learning Objectives                                    |
 | ------------------------------ | --------------------------------- | ------------------------------------------------------ |
 | [Lesson 1](./week-1/lesson.md) | Fundamentals • Semantics          | [Learning objectives](./week-1/learning-objectives.md) |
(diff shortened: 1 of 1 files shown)


SallyMcGrath:
👋 My suggestion here would be instead of removing the note, expand upon it. It's a useful way to approach forms in html (and make this lesson more engaging), to relate it to structuring data more generally. How might you use this opportunity? How is an HTML a structured interaction and how does it help you input data into a database? How do different form inputs constrain data types and why is that useful to do? There's a lot of good conversations to be had here that will set trainees up well for future concepts. So I would not remove the note, but if you think future mentors need more guidance towards that, how about opening a ticket on the boards, or a new PR with the guidance?

Issue #461, opened 2022-07-23
Add context on why writing tests first is useful (as part of TDD)


40thieves:
**Which module(s) and week(s) does this change affect?**
Module(s): JS2
Week(s): 2

**What is the work that needs to be done?**

As discussed here: https://github.com/CodeYourFuture/syllabus/pull/457#discussion_r928154002, we introduce TDD but without adding context on why writing test code before implementation code is a good idea.

We should add this context - either in JS2 W2 or elsewhere.

**Why is this work important to do?**

Adding context would prevent trainees following "dogma" without understanding why. IMO, good engineers bend the rules where necessary, but a good understanding on the reasoning is required for this.

**Additional context**

Seems like @Dedekind561 is planning on adding some more TDD materials.

**Who might need to know about this change?**

@SallyMcGrath 
@CodeYourFuture/global-syllabus

40thieves:
Discussion in global syllabus: we think it's good enough to just add an exercise to reflect on why it's a good idea. Volunteers can steer on their opinions.

40thieves:
I've got a PR to address this comment https://github.com/CodeYourFuture/syllabus/issues/461#issuecomment-1278077074 here: https://github.com/CodeYourFuture/syllabus/pull/491

40thieves:
We think this is good enough for now, pending a wider discussion around TDD. Closing.

PR #491, opened 2022-10-26
Add an exercise to reflect on why TDD is useful

alismith-dev:
## What does this change?

Module: JS2
Week(s): 2

## Description

Adds an exercise that encourages trainees to reflect on why TDD might be useful. The goal here is to give context on what is actually useful about TDD and what is "dogma", hopefully via discussion with trainees and teachers/TAs.

Contributes to https://github.com/CodeYourFuture/syllabus/issues/461. It's possible that we want to extend this as part of wider work to teach testing more explicitly, but we think this might be good enough to cover the points in the ticket for now.

EDIT: I also removed the table on contents at the top of the page, since it's out-of-date and we now have sidebar that is always kept up-to-date.

@CodeYourFuture/global-syllabus 

## Rendered Pages

-----
[View rendered docs/js-core-2/week-2/lesson.md](https://github.com/CodeYourFuture/syllabus/blob/add-tdd-context-exercise/docs/js-core-2/week-2/lesson.md)

### DIFF
--- a/docs/js-core-2/week-2/lesson.md
+++ b/docs/js-core-2/week-2/lesson.md
@@ -12,16 +12,6 @@ import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 import ExerciseSolutions from "@theme/ExerciseSolutions";
 
-## Contents
-
-- [Methods review](#tdd-and-methods)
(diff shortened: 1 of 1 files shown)


gregdyke (approved):

SallyMcGrath (commented):
LGTM

PR #536, opened 2022-11-24
fix: add first step to start dev server

Dedekind561:
## What does this change?

Module: React 
Week(s): 1

## Description

Adds an additional step to get trainees to start the app running with `npm start` 

@40thieves 

## Rendered Pages

-----
[View rendered docs/react/week-1/lesson.md](https://github.com/CodeYourFuture/syllabus/blob/fix/react-week-1-notes/docs/react/week-1/lesson.md)

### DIFF
--- a/docs/react/week-1/lesson.md
+++ b/docs/react/week-1/lesson.md
@@ -388,7 +388,9 @@ function MentorsList() {
 }
 ```
 
-Here we have added a `key` prop to the `li` element. The `key` prop has a special meaning in React because it is used internally to keep track of which element in the list is which. A documentation page explaining in more depth is in the further reading section.
+Here we have added a `key` prop to the `li` element. The `key` prop has a special meaning in React. React uses `key` to keep track of each element in the list. 
+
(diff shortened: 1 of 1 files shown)


docs/react/week-1/lesson.md
```diff
@@ -387,7 +388,7 @@ function MentorsList() {
 }
 ```
 
-Here we have added a `key` prop to the `li` element.  The `key` prop has a special meaning in React because it is used internally to keep track of which element in the list is which. A documentation page explaining in more depth is in the further reading section.
+Here we have added a `key` prop to the `li` element. The `key` prop has a special meaning in React because it is used internally to keep track of which element in the list is which. A documentation page explaining in more depth is in the further reading section.
```
SallyMcGrath:
```suggestion
Here we have added a `key` prop to the `li` element. The `key` prop has a special meaning in React. React uses `key` to keep track of each element in the list. 

The [React documentation](https://reactjs.org/docs/lists-and-keys.html)  and [React learning book](https://beta.reactjs.org/learn/rendering-lists#keeping-list-items-in-order-with-key) explain this in more depth.
```

SallyMcGrath (dismissed):
tiny revision --  broke up the sentences and linked straight to docs

PR #541, opened 2022-11-28
Add context to briefings page.md

kfklein15:
Signed-off-by: KFK <80272258+kfklein15@users.noreply.github.com>

## What does this change?

Module: Final Projects
Week(s): Project Briefings

## Description

 added a phrase to give context of what that pages is there for

## Rendered Pages

-----
[View rendered docs/finalproject/projects.md](https://github.com/CodeYourFuture/syllabus/blob/kfklein15-patch-4/docs/finalproject/projects.md)

### DIFF
--- a/docs/finalproject/projects.md
+++ b/docs/finalproject/projects.md
@@ -4,7 +4,7 @@ title: Project Briefings
 sidebar_label: Project Briefings
 ---
 
-You can find here some examples of projects that you could chose to do or that have been done previously, to give you and your team inspiration.
+These are some examples of projects that could we chosen for your team or projects that have been done previously.
 
 New projects:
 - [A Set](https://docs.google.com/document/d/1_aYRVzw1DJE1K5EadBzZqqh00ZlA4uaXw7izzLJkAIo/edit?usp=sharing)
--- a/docs/finalproject/projects.md
+++ b/docs/finalproject/projects.md
@@ -4,9 +4,9 @@ title: Project Briefings
 sidebar_label: Project Briefings
 ---
 
-These are some examples of projects that could we chosen for your team or projects that have been done previously.
+Here are some examples of previous final projects.
 
-New projects:
+Project briefings:
 - [A Set](https://docs.google.com/document/d/1_aYRVzw1DJE1K5EadBzZqqh00ZlA4uaXw7izzLJkAIo/edit?usp=sharing)
 - [Deskeando](https://docs.google.com/document/d/1GTIKg7UyItzZs-Xq4Kd0yY0WQLHM0L98Hoyr5BoktOg/edit?usp=sharing)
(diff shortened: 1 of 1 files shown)


docs/finalproject/projects.md
```diff
@@ -4,6 +4,8 @@ title: Project Briefings
 sidebar_label: Project Briefings
 ---
 
+You can find here some examples of projects that you could chose to do or that have been done previously, to give you and your team inspiration.
```
SallyMcGrath: Who is the "you" addressed here?

Trainees often ask if they can choose their own final project and team, which they cannot. It would be useful to make the intended audience clear here.

SallyMcGrath: (Also psst it's choose not chose)

SallyMcGrath (commented):
Small discussion point.

docs/finalproject/projects.md
kfklein15: In Glasgow, the trainees were allowed to choose/define their final project. I know it's no longer the case so we can change it.

SallyMcGrath: To complicate this a little, sorry: I'm not against trainees doing this in Glasgow if they have the support and resources to do it effectively. It's not about them being allowed or not; it's about trainees having the ability to organise, plan, and execute a final project. We have had situations in other regions with no final projects being organised.

docs/finalproject/projects.md
```diff
@@ -4,7 +4,7 @@ title: Project Briefings
 sidebar_label: Project Briefings
 ---
 
-You can find here some examples of projects that you could chose to do or that have been done previously, to give you and your team inspiration.
+These are some examples of projects that could we chosen for your team or projects that have been done previously.
```
SallyMcGrath: This doesn't quite make sense as a sentence. It might be helpful to use Grammarly before committing. What about saying instead:

Here are some examples of previous final projects.

SallyMcGrath (changes requested):
Maybe simplify this a little? I think this sentence has got a bit lost in itself.

docs/finalproject/projects.md
kfklein15: But the first list of projects isn't once that has been done, they are possible briefings/projects. So this phrase wouldn't make sense, unless I remove all the possible briefings, which I am happy to do.

SallyMcGrath: I think all those projects have been done.

kfklein15: That makes it easier, indeed. Will amend the PR now and if you can approve it in 5 minutes would be great. Thank you.

SallyMcGrath (commented):
LGTM

SallyMcGrath (approved):

SallyMcGrath:
I approved this on Dec 5 but there was no conversation resolution on the change requests so I guess it got lost and then stale. I've refreshed the branch and approved again. If @esma-g approves it you can merge.

esma-g (approved):
Approved.

Issue #499, opened 2022-10-27
JS1 Week 1 work review


SallyMcGrath:
Which module(s) and week(s) does this change affect?
Module(s): JS1
Week(s): 1

What is the work that needs to be done?
https://github.com/CodeYourFuture/syllabus/issues/494

Why is this work important to do?

Additional context

Who might need to know about this change?

PR #548, opened 2022-12-05
remove extra js function exercise

Dedekind561:
## What does this change?

Module: JS1
Week(s): Week 1

Part of #499

## Description

Removes an extra JS exercise from the syllabus - the same skills are already assessed in subsequent exercises in the coursework.


SallyMcGrath (approved):
Yes I'm happy to ✂️ this. I'm also always confused by the phrase "increase by half" 😕  so let's be free of it!

Issue #570, opened 2023-01-14
Style guide violation, please update


SallyMcGrath:
https://github.com/CodeYourFuture/syllabus/blob/02da63dbe1a20f8df0559074f4dac810e8af62a8/docs/js-core-1/week-4/lesson.md?plain=1#L166

https://github.com/CodeYourFuture/syllabus/blob/02da63dbe1a20f8df0559074f4dac810e8af62a8/docs/js-core-1/week-4/lesson.md?plain=1#L167

Improper use of arr, newArr
https://syllabus.codeyourfuture.io/guides/code-style-guide

Please update to follow our own advice

MohitBansal321:
i can you assgin this issue to me

MohitBansal321:
okay i will replace it with this
const array=[1,2,3];
const newArray=array.concat(4);

MohitBansal321:
i created pull request

SallyMcGrath:
Hi @MohitBansal321 .  You can connect tickets and PRs on GitHub with a hashtag. Type #572  here or on your PR type `fixes` #570

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

MohitBansal321:
#572

MohitBansal321:
I done the changes 
let me know is it okay

MohitBansal321:
@SallyMcGrath 
Hey I am not getting any feedback from you 
kindly respond  me so i can work on another issues here

Issue #569, opened 2023-01-14
Style guide violation, please update


SallyMcGrath:
https://github.com/CodeYourFuture/syllabus/blob/02da63dbe1a20f8df0559074f4dac810e8af62a8/docs/workshops/js-testing-workshop.md?plain=1#L54

Improper use of var

https://syllabus.codeyourfuture.io/guides/code-style-guide

Please update to follow our own advice

MohitBansal321:
what can we use instead of var?

SallyMcGrath:
> what can we use instead of var?

`let`

MohitBansal321:
i created a pull request

MohitBansal321:
I changed all var to let according to style guide @SallyMcGrath
please let me know how is it

PR #572, opened 2023-01-15
Style guide violation 

MohitBansal321:
## What does this change?

Module: js-core-1 , workshops
Week(s):week-4 , js-testing-workshop
#570 #569 
## Description

There is a style guide violation so I changed it according to style guide 

@SallyMcGrath

### DIFF
--- a/docs/js-core-1/week-4/lesson.md
+++ b/docs/js-core-1/week-4/lesson.md
@@ -38,11 +38,11 @@ You can check this by starting a node console in your terminal.
 
 ```sh
 $ node
-> let arr = [1, 2, 3];
+> let arrayOfNumbers = [1, 2, 3];
 undefined
-> arr
+> arrayOfNumbers
 [1, 2, 3]
-> arr.length
+> arrayOfNumbers.length
 3
 ```
 
@@ -150,11 +150,11 @@ _Adds (or concatenates) another value or array to the array._
 
 ```sh
 $ node
-> let arr = [1, 2, 3];
+> let arrayOfNumbers = [1, 2, 3];
 undefined
-> arr.concat(4)
+> arrayOfNumbers.concat(4)
 [1, 2, 3, 4]
-> arr
+> arrayOfNumbers
 [1, 2, 3]
 ```
 
@@ -163,10 +163,10 @@ Did you notice how calling the concat method did not change `arr`? This is becau
 If you want to use the array returned by calling `.concat()` you should store it in a new variable.
 
 ```js
-const temp = [1, 2, 3];
-const newTemp = temp.concat(4);
+const arrayOfNumbers = [1, 2, 3];
+const newReturnedArray = arrayOfNumbers.concat(4);
 
-console.log(newTemp); // logs [1, 2, 3, 4]
+console.log(newReturnedArray); // logs [1, 2, 3, 4]
 ```
 
 ### `.slice()`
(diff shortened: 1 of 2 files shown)


docs/js-core-1/week-4/lesson.md
```diff
@@ -163,10 +163,10 @@ Did you notice how calling the concat method did not change `arr`? This is becau
 If you want to use the array returned by calling `.concat()` you should store it in a new variable.
 
 ```js
-const arr = [1, 2, 3];
-const newArr = arr.concat(4);
+const array = [1, 2, 3];
```
SallyMcGrath: Thanks for this!

I think this still violates the style guide https://syllabus.codeyourfuture.io/guides/code-style-guide#avoid-short-names

Here's the bit I think talks about this:

> Here are some examples of bad names that you should avoid:
> 
> Single letters like x or y
> Abbreviations like evt instead of event
> Generic names like array or string


What are your thoughts?

docs/js-core-1/week-4/lesson.md
```diff
@@ -163,10 +163,10 @@ Did you notice how calling the concat method did not change `arr`? This is becau
...
+const newArray = array.concat(4);
```
SallyMcGrath:
```suggestion
const newArray = array.concat(4);
```

And I think here too, is not following the style guide recommendations.

SallyMcGrath (changes requested):
Thanks so much for this contribution! I've made some notes for you to think about before this is merged. Let me know your thoughts.

docs/js-core-1/week-4/lesson.md
MohitBansal321: okay i will change the array to temp[variable name]  is it okay?
newArray to newTemp

docs/js-core-1/week-4/lesson.md
```diff
@@ -163,10 +163,10 @@ Did you notice how calling the concat method did not change `arr`? This is becau
 If you want to use the array returned by calling `.concat()` you should store it in a new variable.
 
 ```js
-const array = [1, 2, 3];
-const newArray = array.concat(4);
+const temp = [1, 2, 3];
```
SallyMcGrath: > A good variable name quickly explains what it represents to anyone reading the code. Try to describe what the variable is or what it does within in code. ~[ Style Guide](https://syllabus.codeyourfuture.io/guides/code-style-guide#describe-what-the-variable-isdoes)

In this case, we are teaching a beginner how to concatenate two arrays, so it is easy to think that we should say `array` or temp. But really we have to make it clearer to the learner what the variables are **doing**, to make this a richer learning experience for them, as well as being consistent with the rules we set for their coursework.

Here's the context :

> Did you notice how calling the concat method did not change arr? This is because concat, like most array methods, returns a new array; it does not alter the array you called the method on.

> If you want to use the array returned by calling .concat() you should store it in a new variable.

```
const exampleArrayOfNumbers = [1, 2, 3];
const newReturnedArray = exampleArrayOfNumbers.concat(4);

console.log(newReturnedArray); // logs [1, 2, 3, 4]
```

SallyMcGrath (changes requested):
Couple of things here. 

1. I've given a suggestion of what I might do on this ticket. Take a look and have a think.
2. This file actually has more lines referring to arr. Would you mind taking a look at those as well? 

https://github.com/CodeYourFuture/syllabus/blob/2aca2a1d6e31174a023d40137c41e386bb415d5b/docs/js-core-1/week-4/lesson.md

docs/js-core-1/week-4/lesson.md
MohitBansal321: okay got it I will change all the lines which are violating style guide
I just though that you only want to change on particular  lines so I changed that

MohitBansal321: @SallyMcGrath check it

MohitBansal321:
let me know all things are okay
Thank you

PR #576, opened 2023-01-15
improve use of html terminology in html/css lesson (elements rather t…

shieldo:
…han tags)

## What does this change?

Module: HTML/CSS
Week(s): 1

## Description

The section of this lesson on Sectioning Content currently uses the term "tag" when it should more correctly use "element". These terms have already been described earlier in the lesson, but the usage here seems to diverge from these definitions. I've also made some related wording tweaks such as preferring describing content as being "inside" an element rather than "in" an element, and referring to structure of a web page when talking about these page-level sectioning elements.

### DIFF
--- a/docs/html-css/week-1/lesson.md
+++ b/docs/html-css/week-1/lesson.md
@@ -174,7 +174,7 @@ When creating a web page, we should always think about how to structure the cont
 
 ### Sectioning Content
 
-We can group elements with 'sectioning' tags.
+We can apply structure to a web page using 'sectioning' elements.
 
 <img src={require('!file-loader!../assets/webpage-structure.png').default}
 alt="Wireframe of a web page with <header> at the top, <main> at the middle, <footer> at the bottom, and <aside> at the right"/>
@@ -182,11 +182,11 @@ alt="Wireframe of a web page with <header> at the top, <main> at the middle, <fo
 The image above shows a common layout of a web page.
 We can use specific HTML elements for each of these sections.
 
-- We can use a `<header>` tag for the header content
-- We can use a `<footer>` tag for the footer content
-- We can use a `<main>` tag for the main content of the page
-- Sidebar content can go in an `<aside>` tag
-- If there is a list of links, such as in the `<header>`, they can go in a `<nav>` tag
+- We can use a `<header>` element for the header content
+- We can use a `<footer>` element for the footer content
+- We can use a `<main>` element for the main content of the page
+- Sidebar content can go inside an `<aside>` element
+- If there is a list of links, for example within the `<header>` of a page, they can go inside a `<nav>` element
 
 Additionally, we can use `<article>` and `<section>` to divide these sections into more sections.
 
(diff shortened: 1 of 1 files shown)


docs/html-css/week-1/lesson.md
```diff
@@ -174,19 +174,19 @@ When creating a web page, we should always think about how to structure the cont
 
 ### Sectioning Content
 
-We can group elements with 'sectioning' tags.
+We can apply structure to a web page using 'sectioning' elements.
```
SallyMcGrath: I think grouping elements is important to say clearly.  This is because I often teach catchup classes where trainees get a lot of value from discussing and reflecting on the concept of naming and grouping data in html. This is all structuring a page, of course, but the grouping word is useful. 

How about going right to the spec?
```suggestion
We can group elements with sectioning content.  This splits a page up into named sections.
```

docs/html-css/week-1/lesson.md
```diff
@@ -174,19 +174,19 @@ When creating a web page, we should always think about how to structure the cont
...
 
 The image above shows a common layout of a web page.
 We can use specific HTML elements for each of these sections.
 
-- We can use a `<header>` tag for the header content
-- We can use a `<footer>` tag for the footer content
-- We can use a `<main>` tag for the main content of the page
-- Sidebar content can go in an `<aside>` tag
-- If there is a list of links, such as in the `<header>`, they can go in a `<nav>` tag
+- We can use a `<header>` element for the header content
```
SallyMcGrath, line 185: I've got no objection to using element instead of tag. It is the spec after all. We might want to mention tag somewhere to increase their Googlefu.

SallyMcGrath (changes requested):
Thanks, @shieldo 🎉 

Minor notes -- take a look and have a think.

docs/html-css/week-1/lesson.md
shieldo:
```suggestion
We can put groups of content that make up areas of a web page inside 'sectioning' elements. These split the page up into named sections that help to organise the page's structure.
```

I've tried to reflect the way [the part of the HTML spec dealing with sections](https://html.spec.whatwg.org/multipage/sections.html) refers to grouping of content (I'm hoping this is exactly what you're referring to by "grouping elements"). I'm not sure the use of "sectioning content" works here - it has a [particular meaning](https://html.spec.whatwg.org/multipage/dom.html#sectioning-content-2) in the spec that does not include the `main`, `header` and `footer` elements mentioned here in the diagram (though does include `aside` and other groupings like `article` and `section`), and I'm not sure using "sectioning content" in the more vernacular way is clear either.

What do you think @SallyMcGrath ? I do think we're close to something that is hopefully clear and that includes that sense of grouping HTML data within these more macrostructural elements.

docs/html-css/week-1/lesson.md
shieldo: There are mentions of tags elsewhere on the page that are arguably reasonable uses (thinking of the tabular content section). Does that fulfil this concern?

SallyMcGrath: Yes, I think so 👍

SallyMcGrath (approved):
👍 LGTM

PR #582, opened 2023-01-19
CSS Units table improvement

shieldo:
## What does this change?

Module: HTML/ CSS
Week(s): 1

## Description

Improvements to the CSS units table in the Lesson for this week:

1. adding an explicit name column to reflect how developers usually refer to these units in speech (especially to use the name "pixel", which was not previously mentioned in the table)
2. putting `px` first in the list, because two of the other units refer to this unit
3. adding fuller explanation of what a pixel means in CSS, replacing the description of a pixel being 1/96th of an inch, which is over-simple and needed more context (referencing [CSS Level 3 spec](https://www.w3.org/TR/css-values-3/#absolute-lengths))
4. use code markup for examples of numeric CSS values


gregdyke:
I don't know enough about CSS and am unsure how much complexity/precision we want to have in describing a px

shieldo:
Yes, I wasn't sure whether extra colour in the description of a pixel was useful or mostly noise. Given I elided mentioning a visual angle and conveying how a reference pixel takes up larger physical space for screens meant to be viewed from further away (for sake of not blinding with _too_ much science), perhaps it is more the latter. I certainly think the current mention of 1/96th of an inch is misleading and confusing - I helped a trainee in class the other day who was referencing exactly this part of the lesson and wondering why the unit seemed to get bigger when they were increasing zoom size in the browser. So just mentioning that this refers to a "pixel", and how this has a relation to device pixels may be just the right level of detail here?

i.e. it may be we'd want to keep

```md
This used to refer to real physical pixels on screens, but on modern devices with high resolution displays one CSS pixel is usually rendered using a whole multiple of real pixels.
```

and ditch (from my proposal here)

```md
For rendering on screens, a CSS “reference pixel” is defined as the visual angle of one pixel on a device with a pixel density of 96 pixels per inch at a distance of an arm’s length from the reader.
```

docs/html-css/week-1/lesson.md
```diff
@@ -415,30 +415,35 @@ The following table contains some of the most commonly used CSS units.
...
       <th>Relative or absolute</th>
       <th>Description</th>
     </tr>
   </thead>
   <tbody>
+    <tr>
+      <td><code>px</code></td>
+      <td>pixel</td>
+      <td>Absolute</td>
+      <td>This used to refer to real physical pixels on screens, but on modern devices with high resolution displays one CSS pixel is usually rendered using a whole multiple of real pixels. For rendering on screens, a CSS “reference pixel” is defined as the visual angle of one pixel on a device with a pixel density of 96 pixels per inch at a distance of an arm’s length from the reader.</td>
```
SallyMcGrath: So this isn't wrong, but I don't think this is useful to lead with on day 1 module 1. Is it possible you may be falling prey to the [expert blind spot](https://teachtogether.tech/en/index.html#s:memory-becoming-expert)? (the tendency of experts to organize explanation according to the subject’s deep principles rather than being guided by what their learners already know)

Let's instead think about what the trainee needs to know to use a pixel. Do they need to use a pixel? Is a pixel the right thing to lead with? Where would this declaration about pixel rendering fit in a c[oncept map](https://teachtogether.tech/en/index.html#s:conceptmaps) of this lesson?

I really like this video on motivating explanations for trainees https://teachertraining.codeyourfuture.io/content/how-do-i-motivate-explanations-for-students

SallyMcGrath (changes requested):
I have left a general comment, which is basically, rather than doubling down on this, does this table need to be here at all? Could we build something more illuminating instead?

SallyMcGrath:
@shieldo just checking  - are you still working on this?

shieldo:
@SallyMcGrath I think I needed a little distance so I could come back to it with some fresher eyes.

I've simplified the description of the pixel as per above, and also clarified the text about absolute and relative units (the description that absolute units stay the same size when zooming in within a browser wasn't really correct and this text confused at least one trainee in the class I help in).

I think this represents an improvement now, but thanks for the challenge about whether keeping the table is the right approach. It's hard to get a sense of how malleable everything is, but I think I'm gaining that. Given the torrent of different concepts being thrown in during this whole lesson, the first week's lesson of the course, the table here doesn't seem all that out of place,  though it could be an idea to do something like an embedded interactive example where trainees can modify the dimensions of two or three nested elements and see how changing one can modify the rendering of elements with relative sizing but doesn't affect dimensions declared with absolute units? Does that sound like a more useful method if we're trying to explain one of many concepts in this lesson without overburdening?

SallyMcGrath:
@shieldo 

Sorry, I missed this comment at the time (it was during my bereavement and I was a bit all over the place).  These changes are good. Thank you!

SallyMcGrath (approved):

PR #584, opened 2023-01-24
Fix bug with Fundamentals link

textbook:
As a hangover from being a separate site it was an `href` when it should have been a `to`.

Signed-off-by: Jonathan Sharpe <jonathan@codeyourfuture.io>

## What does this change?

Global navigation

## Description

Updates the link to always take you to the root Fundamentals page

- @SallyMcGrath 
- @kfklein15

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -62,7 +62,7 @@ module.exports = {
               label: "Intro To Coding",
             },
             {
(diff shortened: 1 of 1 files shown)


kfklein15 (approved):
Change is being made and should work (not sure I can actually test it). so once merged I'll check again.

PR #585, opened 2023-01-24
Update actions

textbook:
Should make the checkout faster

Signed-off-by: Jonathan Sharpe <jonathan@codeyourfuture.io>

## What does this change?

Updates the actions we use in CI

## Description

Updates `actions/checkout` and `actions/setup-node` to newer versions, and migrates the manual caching to the latter. This gives us less to maintain ourselves, avoids using deprecated features and, as the clone is now shallow, should make the checkout step much faster (3s on this PR's check compared to 25s [here](https://github.com/CodeYourFuture/syllabus/actions/runs/3994916751/jobs/6853179204)).

The installation took slightly longer (26s vs. 13s) but that's from a cold cache, should be just as fast as the old version once it's warm and the time is dominated by the actual build anyway.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,17 +11,11 @@ jobs:
     if: github.event_name != 'push'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
           node-version: "16.x"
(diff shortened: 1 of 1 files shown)


SallyMcGrath (approved):
LGTM 🙇

SallyMcGrath:
(Context for force merge: I put a large protection on the branch as we are working with a new team, so I merged this for you. Though I realise you can do this yourself I thought you probably wouldn't after seeing the 3 approval rule we have right now. )

PR #597, opened 2023-01-25
Update Docusaurus to 2.2.0.

homotechsual:
## What does this change?

Meta change - updates the Docusaurus SSG version to the 2.2.0 stable from Beta 0

## Description

Updates Docusaurus to latest stable due to the stable status future updates in the 2.x series will be minor version upgrades with no breakage potential - though the site was well built in the first instance and didn't require any custom handling.

A random but extensive sampling of 15 pages have been compared to their current versions with no breakages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,9 +8,9 @@
       "name": "codeyourfuture-syllabus",
       "version": "2.0.0",
       "dependencies": {
-        "@docusaurus/core": "^2.0.0-beta.0",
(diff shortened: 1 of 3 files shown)


homotechsual:
Removes an unused import in a page too.

SallyMcGrath:
👋 Thank you @homotechsual !

Another person also committed this and I saw that first, so I will close this now. But thank you so much for the contribution and please add yourself to all-contributors. And come again! 🎉

PR #600, opened 2023-01-30
Fix Search on Syllabus (Remove Algolia)

ChrisOwen101:
## What does this change?

Module: All
Week(s): All

## Description

It was raised on Slack that Search has stopped working because of a reliance on Algolia which powers the search.

This PR fixes this by stopping our reliance on Algolia and changing to a search that runs at build time. It's a tiny bit slower but absolutely does the job.

@SallyMcGrath

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -6,6 +6,7 @@ module.exports = {
   organizationName: "codeyourfuture",
   projectName: "Syllabus",
   onBrokenLinks: "throw",
+  plugins: [require.resolve("docusaurus-lunr-search")],
   presets: [
     [
       "@docusaurus/preset-classic",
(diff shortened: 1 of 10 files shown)


kfklein15 (dismissed):
We need three reviewers right now, so am approving this too. Thanks, Chris!

SallyMcGrath (dismissed):
LGTM -- no objection to lunr -- ta Chris!

SallyMcGrath (approved):
rote approval without check

SallyMcGrath:
@ChrisOwen101 (it's not an emergency because it just doesn't deploy but it will block deploys) this build fails. 

I pulled down this PR before I approved and built it on my local. It was fine here. Can you take a look and see what is failing? I'll revert tomorrow if not so we don't block deploys. Ta!

ChrisOwen101:
Eek sorry! Tested on my end and it worked as well.

I'll take a look when I get home later on this evening.

PR #633, opened 2023-02-21
GH clone issues button

berkeli:
## What does this change?

Module: N/A
Week(s): N/A

## Checklist

- [x] I have read the [contributing guidelines](https://syllabus.codeyourfuture.io/contributing/overview)
- [x] I have checked my spelling and grammar with an [automated tool](https://www.grammarly.com/grammar-check)
- [x] I have previewed my changes to check the [markdown renders](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) as I intend
- [x] I have run my code to check it works
- [x] My changes follow our [Style Guide](https://syllabus.codeyourfuture.io/guides/code-style-guide)

## Description

Adds the functionality to clone issues, as per https://github.com/CodeYourFuture/Changes/issues/2
It is not added to any of the UI elements, simply created to demo how it would be implemented.

GH app repo: https://github.com/berkeli/coursework-helper

@SallyMcGrath @kfklein15

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,24 @@
         "Squirtle",
         "autosave",
         "initialise"
-    ]
+    ],
+    "workbench.colorCustomizations": {
+        "activityBar.activeBackground": "#26394d",
+        "activityBar.background": "#26394d",
+        "activityBar.foreground": "#e7e7e7",
+        "activityBar.inactiveForeground": "#e7e7e799",
+        "activityBarBadge.background": "#ab547e",
+        "activityBarBadge.foreground": "#e7e7e7",
+        "commandCenter.border": "#e7e7e799",
+        "sash.hoverBorder": "#26394d",
+        "statusBar.background": "#15202b",
+        "statusBar.foreground": "#e7e7e7",
+        "statusBarItem.hoverBackground": "#26394d",
+        "statusBarItem.remoteBackground": "#15202b",
+        "statusBarItem.remoteForeground": "#e7e7e7",
+        "titleBar.activeBackground": "#15202b",
+        "titleBar.activeForeground": "#e7e7e7",
+        "titleBar.inactiveBackground": "#15202b99",
+        "titleBar.inactiveForeground": "#e7e7e799"
+    }
 }
--- a/src/components/CloneIssuesButton.jsx
+++ b/src/components/CloneIssuesButton.jsx
@@ -1,9 +1,6 @@
 import React from "react";
 import { useLocation } from "@docusaurus/router";
-
-const API_URL = "http://localhost:3001";
-const BASE_URL = "https://codeyourfuture.io";
-const CLIENT_ID = "Iv1.b36544e8ddc4efc6";
+import { API_URL, BASE_URL, CLIENT_ID } from "../constants";
 
 const CloneIssuesButton = ({ repo, issue = null }) => {
   if (!repo) {
--- a/src/components/CloneIssuesButton.jsx
+++ b/src/components/CloneIssuesButton.jsx
@@ -1,0 +1,63 @@
+import React from "react";
+import { useLocation } from "@docusaurus/router";
+
+const API_URL = "http://localhost:3001";
+const BASE_URL = "https://codeyourfuture.io";
+const CLIENT_ID = "Iv1.b36544e8ddc4efc6";
+
+const CloneIssuesButton = ({ repo, issue = null }) => {
+  if (!repo) {
+    return null;
+  }
+  const token = localStorage.getItem("gh-token");
+  const label = issue ? "Clone" : "Clone All Issues";
+  const location = useLocation();
+  const [loading, setLoading] = React.useState(false);
+  // success is as follows:
+  // {
+  //   issues: 10, // number of issues cloned
+  //   failed: 5, // number of issues that failed to clone
+  // }
+  //
+  const [success, setSuccess] = React.useState({});
+
+  const login = () => {
+    const state = JSON.stringify({
+      prevURL: `${BASE_URL}${location.pathname}`,
+    });
+    window.location.href = `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&state=${state}`;
+  };
+
+  const clone = () => {
+    setLoading(true);
+    const url = issue
+      ? `${API_URL}/clone?repo=${repo}&issue=${issue}`
+      : `${API_URL}/clone?repo=${repo}`;
+    toast.promise(
+      fetch(url, {
+        method: "POST",
+        headers: {
+          Authorization: `Bearer ${token}`,
+        },
+      })
+        .then((res) => res.json())
+        .then((data) => {
+          setLoading(false);
+          setSuccess(data.message);
+        })
+        .catch((err) => {
+          console.log(err);
+        })
+    );
+  };
+
+  return loading ? (
+    <strong>Cloning...</strong>
+  ) : (
+    <button onClick={token ? clone : login} disabled={loading}>
+      {token ? label : "Sign in With Github"}
+    </button>
+  );
+};
+
+export default CloneIssuesButton;
--- a/src/components/CloneIssuesButton.jsx
+++ b/src/components/CloneIssuesButton.jsx
@@ -1,37 +1,17 @@
-import React from "react";
-import { useLocation } from "@docusaurus/router";
+import React, { useState } from "react";
 import useDocusaurusContext from "@docusaurus/useDocusaurusContext";
+import { CircularProgress } from "@material-ui/core";
 
-const SignInButton = () => {
+const CloneButton = ({ module, issue, token, setError }) => {
   const { siteConfig } = useDocusaurusContext();
-  const location = useLocation();
-
-  const qs = new URLSearchParams({
-    client_id: siteConfig.customFields.CLIENT_ID,
-    state: JSON.stringify({
-      prevPath: location.pathname,
-    }),
-  });
-
-  const url = `https://github.com/login/oauth/authorize?${qs}`;
-
-  return (
-    <a href={url} className="button button--primary">
-      Sign In
-    </a>
-  );
-};
-
-const CloneButton = ({ repo, issue, token }) => {
-  const [loading, setLoading] = React.useState(false);
-  const [error, setError] = React.useState(false);
-  const [success, setSuccess] = React.useState(false);
+  const [loading, setLoading] = useState(false);
+  const [success, setSuccess] = useState(false);
 
   const clone = () => {
     setLoading(true);
     const url = issue
-      ? `${API_URL}/clone?repo=${repo}&issue=${issue}`
-      : `${API_URL}/clone?repo=${repo}`;
+      ? `${siteConfig.customFields.API_URL}/github/clone/${module}/${issue}111`
+      : `${siteConfig.customFields.API_URL}/github/clone/${module}`;
 
     fetch(url, {
       method: "POST",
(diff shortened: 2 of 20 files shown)


SallyMcGrath (commented):
🤩 

(Will look tomorrow/Thurs.)

SallyMcGrath:
@40thieves this is for https://github.com/CodeYourFuture/Changes/issues/2

src/pages/auth.js
```diff
@@ -0,0 +1,74 @@
...
+  let redirectURL = "/";
+
+  if (stateString) {
+    const state = JSON.parse(stateString);
+    redirectURL = state.prevPath;
+  }
+
+  // if no code is present, redirect to home page in 3 seconds
+  if (!params.get("code")) {
+    setTimeout(() => {}, 3000);
```
alismith-dev: I think this timeout is not being used, it can be removed.

src/pages/auth.js
```diff
@@ -0,0 +1,74 @@
...
+
+    React.useEffect(() => {
```
alismith-dev: Nesting hooks inside `if` statements is technically not allowed in React: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level. If we had the React linting rules enabled in the syllabus (and perhaps we should!) then this would be marked as an error.

I think probably the best way to avoid this particular case is to extract out the redirect countdown to a separate component (e.g. `RedirectCountdown`), then conditionally render that component.

src/pages/auth.js
```diff
@@ -0,0 +1,74 @@
...
+
+const Auth = () => {
+  const location = useLocation();
+  const params = new URLSearchParams(location.search);
+  const [count, setCount] = React.useState(3);
+  const [loading, setLoading] = React.useState(true);
+
+  const stateString = params.get("state");
+
+  const redirect = (url) => <Redirect to={url} />;
```
alismith-dev: Could you explain your reasoning on why we want a "sub-component" like this? It seems just as simple to render `<Redirect to={redirectURL} />`, but perhaps I'm missing something?

src/pages/auth.js
alismith-dev: I think this would also help clarify some of the states in this component: I was initially pretty confused why a `Auth` component needed a `count` state.

src/constants.js
```diff
@@ -0,0 +1,4 @@
+// URL for the GitHub App API (coursework-helper)
+export const API_URL = "http://localhost:3001";
+// Client ID for the GitHub App (coursework-helper)
+export const CLIENT_ID = "Iv1.b36544e8ddc4efc6";
```
alismith-dev: From what I can tell from the Github docs, this is a secret so shouldn't be exposed in the source: https://docs.github.com/en/rest/guides/basics-of-authentication?apiVersion=2022-11-28#accepting-user-authorization

I think we could look into installing [`dotenv`](https://www.npmjs.com/package/dotenv) into the syllabus to read them from the environment.

src/constants.js
```diff
@@ -0,0 +1,4 @@
+// URL for the GitHub App API (coursework-helper)
+export const API_URL = "http://localhost:3001";
```
alismith-dev: I don't have the coursework-helper installed so `localhost:3001` doesn't work, but switching this to https://coursework-helper.onrender.com/ does 🙂 

This is another good argument for using env vars: setting this URL based on env vars gives us more flexibility.

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+    const state = JSON.stringify({
+      prevPath: `${location.pathname}`,
+    });
+    window.open(
+      `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&state=${state}`,
+      "_self"
+    );
+  };
+
+  return <button onClick={login}>Sign In</button>;
```
alismith-dev: Since we are triggering a page navigation here, it is best to use `<a>` instead of `<button>` and just pass the URL to `href`. This also means that the `_self` argument is unnecessary: it is the default for `<a>`.

To style it to look like a button, you can use the underlying CSS framework in Docusaurus: https://infima.dev/docs/components/button/

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+import { API_URL, CLIENT_ID } from "../constants";
+
+const SignInButton = () => {
+  const location = useLocation();
+  const login = () => {
+    const state = JSON.stringify({
+      prevPath: `${location.pathname}`,
+    });
+    window.open(
+      `https://github.com/login/oauth/authorize?client_id=${CLIENT_ID}&state=${state}`,
```
alismith-dev: This is really nitpicky, but if you wanted to lean on JS features for building URLs then you could do something like this:

```js
const qs = new URLSearchParams({
  client_id: CLIENT_ID,
  state: JSON.stringify({
    prevPath: location.pathname,
  }),
});
const url = `https://github.com/login/oauth/authorize?${qs}`;
```

It's perhaps not worth it for this case, but if we were adding loads of query params programatically then `URLSearchParams` is helpful.

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+      "_self"
+    );
+  };
+
+  return <button onClick={login}>Sign In</button>;
+};
+
+const CloneButton = ({ repo, issue = null, token = null }) => {
```
alismith-dev: Given that `CloneButton` shouldn't ever be rendered without `token`, I would lean towards not defaulting it to null.

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+  if (!repo || !token) {
+    return null;
+  }
```
alismith-dev: This is another case where we're technically using a hook conditionally: it is possible for `useState` to not run.

I think I would throw an error if these props aren't provided.

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+  const [state, setState] = React.useState({
+    loading: false,
+    error: false,
+    result: null,
+  });
```
alismith-dev: Using an object in state like this means that you always have to spread the old state when changing it. The React hooks FAQ recommends against it: https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables

I would generally lean towards either: separating these into their own state variables, or switching to `useReducer` which is a bit more explicit.

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+        setState({ loading: false, error: true, result: err });
+        console.log(err);
+      });
+  };
+
+  if (state.loading) return <strong>Cloning...</strong>;
+  if (!state.loading && !state.error && state.result) {
+    return <strong>Cloned!</strong>;
+  }
+  if (state.error) return <strong>Something went wrong!</strong>;
```
alismith-dev: A couple of things here:

- `!state.loading` will always be true in the first condition, since we always return early on the line above, so I think it might be unnecessary?
- If you switched these two conditions around then I think `!state.error` would also always be true so again would be unnecessary

This is a pattern I like to follow in my conditionals: try to return early for loading & error states, so that you know that it will be in a success state afterwards.

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <button onClick={clone} disabled={state.loading}>
```
alismith-dev: Similar to my comment above: this button will only be rendered if `state.loading` is false, so I think the `disabled` attribute is probably unnecessary.

However... I've just realised that you are rendering different elements during the loading state. From a user experience perspective, the button will "flicker" between different visual states while loading. I wonder if it would be better to keep the `disabled` attribute and switch the text within the button between e.g. `{label}` and `"Cloning..."`

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+      {label}
+    </button>
+  );
+};
+
+export default (props) => {
```
alismith-dev: When you export components like this, they actually don't get a useful "name" attached to them. You can see this in the React dev tools:

Screenshot 2023-03-02 at 16 47 50

Whereas if I give the function a name, then it looks like this:

Screenshot 2023-03-02 at 16 51 43

I find it helps with debugging.

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,69 @@
...
+  if (!repo || !token) {
+    return null;
+  }
+  const [state, setState] = React.useState({
+    loading: false,
+    error: false,
+    result: null,
+  });
+
+  const label = issue ? "Clone" : "Clone All Issues";
```
alismith-dev: Small nitpick: I like to keep code that decides how to render something close to where it actually appears in the JSX (if possible).

So I would consider moving this directly into the `<button>` on lines 60 - 62.

src/components/CloneIssuesButton.jsx
alismith-dev: This is perhaps more of an "UX" issue, but the error message here doesn't really help the user in this situation. Ideally if there's an error state, we should try to tell the user how to fix it.

I suspect the best option here would be to say something like "Refresh the page and try again", but maybe @SallyMcGrath has opinions on that.

alismith-dev (commented):
Amazing work @berkeli!! 👏🏻🎆 I've left quite a few comments, but they're mostly pretty minor stuff. Well done 🙂

src/pages/auth.js
berkeli: Yeah, thanks.

src/pages/auth.js
berkeli: No, you are not missing anything. I think I did it that way because it was going to be called in few places. But extracting to a separate component makes more sense

src/constants.js
berkeli: client ID is not a secret (it is a URL param for github auth), but adding env variables makes sense.

src/constants.js
berkeli: Makes sense, thanks

src/components/CloneIssuesButton.jsx
berkeli: Updated, thanks.

src/components/CloneIssuesButton.jsx
berkeli: Thanks, I was thinking of that but completely forgot when refactoring.

src/components/CloneIssuesButton.jsx
alismith-dev: I think this comment hasn't been fully addressed?

The `state` variable doesn't exist any more, so can be removed.

But I think I would be tempted to further refactor this to address the "flicker" point above, probably something like:

```js
if (error) /* error message */
if (success) /* success message */

<button onClick={clone} disabled={loading}>
  {loading ? "Loading&hellip;" : issue ? "Clone" : "Clone All Issues"}
</button>
```

src/components/CloneIssuesButton.jsx
alismith-dev: Thanks for addressing this comment.

However... I think "check the console" is perhaps not the most actionable error message: it doesn't really help the user with what they have to do to fix the problem. It's unlikely that the console error would be very useful to help them fix it (particularly for trainees at the beginning of the course).

Hence why I suggested something like "Refresh the page and try again": this hopefully helps the user actually fix the problem. I'd perhaps even consider something like: "Refresh the page and try again. If you're still having problems, then contact your Course Leaders".

src/components/CloneIssuesButton.jsx
```diff
@@ -0,0 +1,71 @@
...
+      })
+      .catch((err) => {
+        setLoading(false);
+        setError(err);
+        console.log(err);
+      });
+  };
+
+  if (error) {
+    return <strong>Something went wrong! See console for details.</strong>;
```
alismith-dev: My worry here is that `<strong>` is perhaps not obvious enough for users to notice? Perhaps we could use the [Alert](https://infima.dev/docs/components/alert) styling?

.vscode/settings.json
```diff
@@ -4,5 +4,24 @@
         "Squirtle",
         "autosave",
         "initialise"
-    ]
+    ],
+    "workbench.colorCustomizations": {
```
alismith-dev, line 8: I think this change has been added by accident?

alismith-dev (commented):
Great work @berkeli, thanks a lot for making those changes!

I've left a couple more minor comments, mostly focused on the "UX", but I think this is almost there now 👍🏻

SallyMcGrath:
closing this as it's moved to https://github.com/CodeYourFuture/curriculum

PR #646, opened 2023-03-08
change deploy workflow to use main instead of master

gregdyke:
After the change of the merge branch from master to main, the deploy stopped working. This maybe fixes it?


textbook (approved):
Yep that should do it

PR #659, opened 2023-03-29
Curriculum rewrite/js1 week 1

Dedekind561:
## What does this change?

Module:
Week(s):

## Checklist

- [ ] I have read the [contributing guidelines](https://syllabus.codeyourfuture.io/contributing/overview)
- [ ] I have checked my spelling and grammar with an [automated tool](https://www.grammarly.com/grammar-check)
- [ ] I have previewed my changes to check the [markdown renders](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) as I intend
- [ ] I have run my code to check it works
- [ ] My changes follow our [Style Guide](https://syllabus.codeyourfuture.io/guides/code-style-guide)


docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+As with `Math.round` we need to use the syntax for calling this function.
+So we can add parentheses after the function name to indicate we're calling this function:
+
+```js title="example.js"
+console.log("hello there!");
+```
+
+Now if we our terminal, we should see the string "hello there!" logged out in the terminal.
+
+### Saving function output
```
gregdyke, line 317: I think trainees fail to generalise from this to understand that expressions of arbitary complexity can go in any place where an expression can go (e.g. the RHS of an assignment)

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+So this new problem can be expressed as a question:
+
+> #### ❓ Problem
+>
+> "How can we inspect the data in our program during runtime?"
+> In other words: "how can I print values to the terminal when my program is being executed?"
+
+In order to do this, we can use another function.
+
+### Printing to the terminal
```
gregdyke, line 295: "printing" ? outputting? displaying?

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+`Math.round(10.3)` is a call expression: it can be thought of as saying apply the set of instructions for `Math.round` to the number `10.3`.
+
+If we type `Math.round(10.3)` then again we get the result 10. So once more we can say that `Math.round(10.3)` evaluates to 10.
+
+:::note
+Important to note that `Math.round` is **implementation opaque**. **Implementation opaque** means we can’t read the set of instructions for `Math.round`. This is because `Math.round` was authored by the developers that created NodeJS.
+:::
+
+### Running scripts
```
gregdyke, line 262: I *really* like this motivation, but it feels maybe that it misses the point that expression return values are thrown away unless something is done with them. So it's not that we can't see them, but also that we are doing nothing with them.

One trainee had their code littered with
```
sum(x,y);
console.log(sum(x,y));
```
because they didn't realise that the first value was thrown away (as opposed to "saved somewhere to be reused later")

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+Calling a function means telling the computer to read the function's instructions and carry out its instructions. When calling a function we can also pass inputs to the function.
+
+:::
+
+`Math.round(10.3)` is a call expression: it can be thought of as saying apply the set of instructions for `Math.round` to the number `10.3`.
+
+If we type `Math.round(10.3)` then again we get the result 10. So once more we can say that `Math.round(10.3)` evaluates to 10.
+
+:::note
+Important to note that `Math.round` is **implementation opaque**. **Implementation opaque** means we can’t read the set of instructions for `Math.round`. This is because `Math.round` was authored by the developers that created NodeJS.
```
gregdyke, line 259: I would call it "abstraction" or "encapsulation". It's not so much that we can't go and find out what it does, but also that we don't want to find out (like http - nobody reaaaally wants to know how http works under the hood and fortunately we don't have to)

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+![use-greeting-variable](./assets/use-greeting-variable.png)
+
+:::note
+
+In the example above, we're using backticks to create a template literal. With template literals, we can insert expressions into strings to produce new strings. Any time we want to reference a variable inside a template literal we use a dollar sign and a set of curly braces. The variable name itself is then placed inside the curly braces. The value this variable holds is then placed inside the string.
+
+:::
+
+### Reusing instructions
```
gregdyke, line 208: same as for variables, I'm not sure that there's a point in introducing functions until we can motivate what happens when you don't have reusable instrucitons

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+```bash
+32 / 10
+```
+
+```bash
+3 ** 4
+```
+
+### Saving expressions
```
gregdyke, line 151: Intuitively, I feel like "store" and "save" are different metaphors. I would probably stick to "store", as "saving" is what I do with files. Or maybe "set aside for later?"

I'm concerned that we don't know why we would want to reuse a value - maybe there we can start writing really complex expressions and realise that we maybe want to describe intermediate stages?

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+Often in programming we may want to reuse a particular expression in our program. In this case, we need a way to refer back to a particular expression. Let’s consider the following string:
```
gregdyke, line 153: reuse a _value_, not reuse an _expression_

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+**Data** consists of information in different forms: for example, numbers and text could be different forms of data. The data in a executing program is sometimes referred to as the **state**.
+:::
+
+:::note
+**Operations** are anything that take data and modify existing data or create new data from the current data in the program.
+:::
+
+In other words, any computer program will specify what data we've got and how it will be processed.
+
+### Classifying data
```
gregdyke, line 76: classifying is what Machine learning does - can we just say "Types of data"

gregdyke (commented):
I think this is absolutely great. Way better than what we have now!!

I have a bunch of comments, but all are at the nitpick/"I'm concerned that" level

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Understanding the computer 💻
+
+Modern computers are highly complex devices. It would be too difficult to list all the components that make up a modern computer. Instead, we can form a useful abstraction to make sense of what a computer is and what it does. An **abstraction** is a simplification that we use to focus on certain key aspects of a particular thing. In other words, we can form an **abstraction** of a computer that can give us some limited understanding of it. At the same time, we don't need to delve into the complex inner workings of the computer itself. So we begin with the following abstraction:
+
+:::note
+
+Abstraction: A computer is a device used to store and process data. Computers have the capacity to store data through their memory and the ability to change this data through its processing capacity.
+
+:::
```
alismith-dev, line 42: I would consider cutting this section, I'm not sure bringing in the jargon brings much to the explanation. I think the "Interacting with the computer" section can stand alone without it

Having said that I would be in favour of discussing what abstraction means at some point, given that it's a fairly commonly used phrase.

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Interacting with the computer
+
+As a user, we don't need to understand the complexities of the computer's inner workings. As users we need to harness the computer's memory and processing ability to perform certain tasks. For this purpose, we can issue commands or instructions to the computer for it to carry out some tasks. A computer terminal is an application where users can enter commands to a computer. On a Mac, the terminal looks like this:
+
+![terminal](./assets/terminal.png)
```
alismith-dev, line 48: These examples note that this is what it looks like on Mac specifically - but few trainees will have Macs. Perhaps it's also worth saying that trainees will likely have a different terminal prompt? There's a risk of introducing doubt in trainees if we're too specific about how things should look (particularly this early in the course).

I'd suggest that we instead say something like "The terminal looks something like this:"

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+In other words, any computer program will specify what data we've got and how it will be processed.
+
+### Classifying data
+
+In this course, we’re going to focus on the JavaScript programming language.
+Any programming language will have different forms of data that we call **data types**.
+
+:::note
+A **data type** is a grouping of data with some particular properties.
```
alismith-dev, line 82: I think types are much better understood from a set of concrete examples rather than an abstract description. However... there's a danger of going into too much depth if we just list all the types at this point in the lesson.

alismith-dev: So this got me asking what the value of this section is? I'm not sure I see how it fits into the explanation of how computers work/how expressions implement this.

This got me wondering about whether we can cover types in another way, something like this:

1. This is a number
2. This is how number expressions work
3. Numbers are just one type, this is a string
4. This is how string expressions work

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+:::info
+An **expression** is a value or any valid combination of values and symbols that results in a single value.  
+We say that expressions **evaluate** to a single value.  
+So we say that `10 + 32` **evaluates** to the value `42`.
+:::
+
+### Evaluating expressions
+
+So we can look at an expression like `36 * 45` and ask: what does this expression evaluate to? If we understand what the `*` operator represents ( in this case multiplication ) and if we understand some basic rules of arithmetic we can evaluate this expression.
```
alismith-dev, line 121: Similar to Greg's points: I think we should build some motivation to _why_ we want to ask "what does this expression evaluate to?".

My answer to this would be to touch back to the opening section about how all programs are about manipulating data - i.e. all programs must be built up from thousands of expressions manipulating data.

I do think it would be worth explicitly tying this section to the opening section about how computers work.

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+We say that expressions **evaluate** to a single value.  
+So we say that `10 + 32` **evaluates** to the value `42`.
+:::
+
+### Evaluating expressions
+
+So we can look at an expression like `36 * 45` and ask: what does this expression evaluate to? If we understand what the `*` operator represents ( in this case multiplication ) and if we understand some basic rules of arithmetic we can evaluate this expression.  
+However, we can use computers to evaluate expressions. We can use NodeJS, a JavaScript runtime, an application that runs JavaScript programs.
+
+Node can be used in a number of different ways; however, one particular mode is the REPL. The Node REPL enables us to input JavaScript expressions which are then evaluated by NodeJS and printed to the terminal. The REPL will print out the result of this evaluation.
```
alismith-dev, line 124: Again, what is the trainee's motivation for doing this?

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+:::note
+
+ℹ️ Definition: REPL is a special type of program that stands for
+
+**Read**  
+**Evaluate**  
+**Print**  
+**Loop**
```
alismith-dev, line 133: If I were teaching this, I would probably do some kind of demo to explain what each of these parts meant. I'm struggling to write it up as an exercise, so perhaps that's a sign that it's actually not useful to explain.

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+```bash
+10 + 32
+```
+
+```bash
+32 / 10
+```
+
+```bash
+3 ** 4
```
alismith-dev, line 148: Is this intentionally using the [Exponentiation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Exponentiation) operator? I'm not sure I've ever actually used that in a real program - perhaps we should use multiplication instead?

docs/new-curriculum/js-core-1/week-1/notes.md
alismith-dev: > would probably stick to "store", as "saving" is what I do with files

I do agree that I would generally stick to "store".

I think we have to accept that volunteers are likely to use mixed metaphors when discussing this topic, but being consistent in the text seems like a good thing to do.

> maybe "set aside for later?"

This feels quite unwieldy - I'm not sure I would ever actually say it out-loud - so I'd lean towards something more succinct even if it's not quite the right metaphor.

---

> we don't know why we would want to reuse a value

I had the same thought.

Looking ahead to the `console.log` section, I like how it is framed as a problem. It occurred to me that we could pose a similar problem with expressions/variables too: "Once we've run an expression we can't reference the result again in the future (without typing it out), how can we get the computer to remember a result to use again the future?"

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+### Saving expressions
+
+Often in programming we may want to reuse a particular expression in our program. In this case, we need a way to refer back to a particular expression. Let’s consider the following string:
+
+```js
+"Hello there";
+```
```
alismith-dev, line 157: This is the first usage of a string expression.

I suggested above that we perhaps should cover number expressions first, then string expressions. But I now wonder if perhaps it would be useful to go through number expressions **and** variables, then come back to strings?

I think I would prefer to order it so that the concepts of expressions and variables come before the specifics of numbers and strings?

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+We can break down the different syntactic elements of this variable declaration:
+
+- `const` is a keyword used to indicate we're creating a variable.
+- `greeting` is the variable name - like the name of the label for our piece of data.
+- `=` this is the assignment operator. It means assign to the label `greeting` the expression on the right hand side.
+- `"Hello there"` - this is the value we're assigning to the label `greeting`.
+
+We can type this variable declaration into the REPL:
+
+![greeting-variable-declaration](./assets/greeting-variable-declaration.png)
```
alismith-dev, line 192: This shows the variable declaration returning `undefined`. I feel pretty confident that it'll come up as a question.

I had to [double check this](https://stackoverflow.com/questions/49470717/mac-terminal-node-repl-output-for-variable-declarations), but this is because using `const` to declare the variable makes it a statement, not an expression, so it returns `undefined`.

This does not seem like it is easy to explain, so I guess we should just tell trainees to ignore it? Doesn't feel very satisfying to me.

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+Now we can refer to the label `greeting` again in the REPL:
+
+![reference-greeting](./assets/reference-greeting.png)
+
+Now we have the `greeting` variable stored in memory we can reuse it to build more expressions:
+
+![use-greeting-variable](./assets/use-greeting-variable.png)
```
alismith-dev, line 200: I realise it's explained below, but this is another place we're using new syntax to explain a concept.

Again coming back to my point about how we order number & string expressions, I think this supports my idea of only using number expressions in the variable section, then cover strings separately. You could switch this example to something like:

```
$ const fortyTwo = 10 + 32;
undefined
$ fortyTwo;
42
$ fortyTwo + 58;
100
```

docs/new-curriculum/js-core-1/week-1/notes.md
alismith-dev: I think this section does give some motivation on why functions are useful: there's the discussion of rounding numbers repeatedly.

Perhaps reframing this as a question for trainees would make it more obvious?

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+Given the number `10.3`, we could ask:
+
+> 🤔 "What is the nearest whole number to `10.3`?"
+
+The process of finding the nearest whole number to 10.3 is called **rounding**. So we could rephrase the previous question to be:
+
+> 🤔 "What does the number `10.3` **round** to?”
+
+Again we can use Node REPL in order to round the number 10.3 to the nearest whole number. However, in this case, there is no arithmetical operator for rounding the number `10.3` in JavaScript.
+However, we will often want to carry out some kind of operation or series of operations repeatedly. For example, we will often want to round numbers again and again. So in this case we can use a **function**.
```
alismith-dev, line 220: > For example, we will often want to round numbers again and again

To me this implies that we're rounding the same number multiple times. I wonder if just changing it to "different numbers" would help?

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+
+:::note
+
+Definition: A function is a reusable set of instructions.
```
alismith-dev, line 224: I do like (and use) the metaphor of instructions often, but would it be worth making it explicit that "instructions" are the same as expressions? (I realise this is not technically correct, since it's also statements).

This helps tie the sections together?

docs/new-curriculum/js-core-1/week-1/notes.md
alismith-dev: I think we should just cut this paragraph: it feels like something a teacher can cover themselves if it comes up as a question.

There's other things that I would prefer to cover over this.

docs/new-curriculum/js-core-1/week-1/notes.md
alismith-dev: I do think mentioning "printing" is a commonly used phrase, so worth explaining what it means. But yeah I'd frame it in terms of "displaying something on screen".

This might also help explain the point: the computer is still doing something, but we can't tell unless we tell it to display something.

docs/new-curriculum/js-core-1/week-1/notes.md
```diff
@@ -0,0 +1,349 @@
...
+`console.log` will allow us to print expressions to the console. We also say that we log values to the console. This is very useful for complex programs when we need to inspect the result of certain expressions.
+Let's see how to use `console.log`
+
+In a file called `example.js`, we can write name of the function `console.log`.
+
+```js title="example.js"
+console.log;
+```
+
+However, if we run this file with Node, we won't be able to see anything in the terminal.
```
alismith-dev, line 307: I think I would drop these bits: it's really just repeating the points from the functions intro section. I think I would just go straight into the comparison with calling `Math.round()`.

docs/new-curriculum/js-core-1/week-1/notes.md
alismith-dev: I agree with this point, but perhaps more concretely this section has no explanation that calling a function is itself an expression, and so just like other expressions, the output can be "stored" as a variable. I do think this is an important thing to be explicit about.

I think this would also clarify the point about `console.log("hello world")` returning `undefined` - this is what the expression evaluates to.

alismith-dev (commented):
Thanks for taking this on @Dedekind561, I think this is a good start 👍🏻 

---

I've made the assumption that you're looking for input on specific wording, and not just on the concepts/ordering. Feel free to ignore if that's not the case.

---

I worry that the writing style is overly technical, so much so that I would expect a lot of volunteers to soften it while teaching or possibly suggest rewriting it altogether. I also think the writing style is also quite formal and not very "friendly". When I write content for the syllabus, I generally try to capture what I would actually say out-loud in a lesson, which I find leads to a more "authentic" style, and I wonder if the syllabus would benefit from a similar approach?

---

I feel that this is missing explicit exercises, which has the danger of turning into a long lecture in the hands of a poor teacher. What's your plan for making it more interactive?

PR #665, opened 2023-04-06
remove old db content

Dedekind561:
Removes old db content leaving behind shiny new Keith B version


SallyMcGrath (dismissed):
I'm happy with this broadly speaking.

I note: there's no coursework view and those links will be dead.

PR #713, opened 2023-07-06
Add minutes from 2023-07-06

illicitonion:


docs/contributing/minutes.md
```diff
@@ -4,6 +4,81 @@ title: Global Syllabus Team Minutes
...
+Consistent feedback we get from employeers:
+* Graduates are too eager to fork repos.
+* Graduates are confused about forks vs branches.
+* Graduates always add all of their files and don't pay much attention to content they shouldn't be committing.
+
+Ideally we would be selling git as a safety net which _reduces_ fear.
+
+We value about the current module that we introduce different concepts gradually, e.g. we don't cover merge conflicts early.
+
+Ali would like to remove forking early on
```
alismith-dev: Maybe to expand on this: IME most companies just give write access to a repo (possibly with controls over modifying `main`). I never forked until I made an OSS contribution.

So if there was some way we could have trainee's work on their own github profile, but without them having to fork and make a PR across repos, then we'd simplify our coursework flow greatly.

If we could do this, it doesn't exclude us from teaching forking later.

docs/contributing/minutes.md
```diff
@@ -4,6 +4,81 @@ title: Global Syllabus Team Minutes
...
+
+Daniel would like to have the git teaching independently of code - have our trainees learn by writing stories or something, so we can divorce "learning git stuff" from the fear of breaking/losing their code.
+
+#### HYF Next Steps
```
alismith-dev: Is this supposed to be here?

docs/contributing/minutes.md
alismith-dev: Perhaps that could be shorted to:

```suggestion
Ali would like to remove forking early on as it would reduce the complexity of the coursework workflow.
```

alismith-dev (commented):
I think there's a stray header? Otherwise I expanded a bit on a point about git - up to you whether you take my suggestion or not.

docs/contributing/minutes.md
```diff
@@ -4,6 +4,81 @@ title: Global Syllabus Team Minutes
...
+
+Ideally we would be selling git as a safety net which _reduces_ fear.
+
+We value about the current module that we introduce different concepts gradually, e.g. we don't cover merge conflicts early.
+
+Ali would like to remove forking early on as it would reduce the complexity of the coursework workflow.
+
+Daniel would like to have the git teaching independently of code - have our trainees learn by writing stories or something, so we can divorce "learning git stuff" from the fear of breaking/losing their code.
+
+#### HYF Next Steps
```
SallyMcGrath:
```suggestion

```

SallyMcGrath (changes requested):
rm the header for you

SallyMcGrath (approved):
LGTM

PR #734, opened 2023-11-04
Note that syllabus team minutes have moved

illicitonion:
Otherwise I am 100% going to make PRs to the old location and it will be annoying for everyone.


alismith-dev (approved):
LGTM, thanks. I would definitely have done that too!

